PR #12, opened 2026-05-21
Add cmd docker and helm for orchestrator

jessicaochen:
## What does this PR do?

Add the needed framing so the accelerator orchestrator can run in a kubernetes cluster

## How was this tested?

- [ x ] Manual testing performed - I deployed it and it runs

## Checklist


Makefile
```diff
@@ -3,6 +3,8 @@
 PROJECT_NAME ?= llm-d-rl-time-slicing
 REGISTRY ?= ghcr.io/llm-d
 IMAGE ?= $(REGISTRY)/$(PROJECT_NAME)
+ORCHESTRATOR_IMAGE ?= $(REGISTRY)/$(PROJECT_NAME)/acceleratororchestrator
```
Edwinhr716, line 6: Why not use `$(IMAGE)` instead of `$(REGISTRY)/$(PROJECT_NAME)`?

jessicaochen: I could but I think the normal understaning is that IMAGE is the actual image name. We could switch it to something like IMAGE_PATH if we want it to represent the shared path instead.

Edwinhr716: Makes sense, I'll update my PR to use `$(REGISTRY)/$(PROJECT_NAME)` instead

droot (approved):

PR #17, opened 2026-06-01
Add the internal state storage for orchestrator

jessicaochen:
Concurrency compatible to allow for multi-thread usage.

## How was this tested?

## Checklist


pkg/accelerator-orchestrator/store/group.go
```diff
@@ -0,0 +1,141 @@
...
+	pb "github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/accelerator-orchestrator/api/v1alpha1"
+)
+
+// Group represents the in-memory and persistent state of a time-slice group.
+type Group struct {
+	mu             sync.RWMutex
+	id             string
+	nodes          []string
+	lockingJob     string // job_id holding the lock, empty if none
+	activeJob      string // job_id whose context is active, empty if none
```
BogdanAtWork: When would activeJob != lockingJob?

pkg/accelerator-orchestrator/store/job.go
```diff
@@ -0,0 +1,83 @@
...
+	j.pods = append([]string(nil), pods...)
+}
+
+func (j *Job) ContextState() map[string]pb.SnapshotAgentJobState_State {
+	j.mu.RLock()
+	defer j.mu.RUnlock()
+	if j.contextState == nil {
+		return nil
+	}
+	res := make(map[string]pb.SnapshotAgentJobState_State)
```
BogdanAtWork: `maps.Clone`? same below.

pkg/accelerator-orchestrator/store/job.go
```diff
@@ -0,0 +1,83 @@
...
+		res[k] = v
+	}
+	return res
+}
+
+func (j *Job) SetContextState(cs map[string]pb.SnapshotAgentJobState_State) {
+	j.mu.Lock()
+	defer j.mu.Unlock()
+	if cs == nil {
+		j.contextState = nil
```
BogdanAtWork: Is there any semantic difference between a nil context and an empty one? Can we just converge on an empty one thus eliminating potential panics downstream?

pkg/accelerator-orchestrator/store/job_store.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// NewJobStore creates a new JobStore.
+func NewJobStore() *JobStore {
+	return &JobStore{
+		jobs: make(map[string]*Job),
+	}
+}
+
+func makeJobKey(jobID, groupID string) string {
```
BogdanAtWork: somewhat of a nit, but we're using (groupId, jobId) ordering in the configmap/mem store, but (jobId, groupId) ordering here. since they're both strings, we should try and be consistent in ordering.

pkg/accelerator-orchestrator/store/group.go
jessicaochen: When the locking job yields but there is no waiting jobs. This is an optimization in the design so that when there is only 1 RL job, we do not bother offloading and reloading its context.

I will fill out the docs a bit more for this.

pkg/accelerator-orchestrator/store/job.go
jessicaochen: good idea, done.

pkg/accelerator-orchestrator/store/configmap_lockstore.go
```diff
@@ -0,0 +1,126 @@
...
+		},
+		Data: make(map[string]string),
+	}
+	created, err := s.client.CoreV1().ConfigMaps(Namespace).Create(ctx, cm, metav1.CreateOptions{})
+	if err == nil {
+		return created, nil
+	}
+	// If it was created concurrently, try to get it again
+	if apierrors.IsAlreadyExists(err) {
+		return s.client.CoreV1().ConfigMaps(Namespace).Get(ctx, ConfigMapName, metav1.GetOptions{})
```
aishukamal: This could lead to a non-IsNotFound error again. I think you probably want lines 34-40 in a function and call that before and here.

pkg/accelerator-orchestrator/store/job.go
jessicaochen: sounds good, did convergence.

pkg/accelerator-orchestrator/store/job_store.go
jessicaochen: done.

pkg/accelerator-orchestrator/store/configmap_lockstore_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"errors"
+	"testing"
+
+	"github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/accelerator-orchestrator/store"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestConfigMapLockStore_LockAndUnlock(t *testing.T) {
```
aishukamal, line 19: Can we also test the RetryOnConflict behavior for both Lock() and Unlock()?

pkg/accelerator-orchestrator/store/configmap_lockstore.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	if cm.Data == nil {
+		return "", nil
+	}
+	return cm.Data[groupID], nil
+}
+
+// Lock persistently sets the job_id holding the lock for the group.
+func (s *ConfigMapLockStore) Lock(ctx context.Context, groupID, jobID string) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
aishukamal, line 81: I'm new to the k8s world and apologies if this is a naive question. Does RetryOnConflict retry on a specific error code only and is ConfigMaps.Update() guaranteed to return that error code only when a version conflict occurs? Just trying to make sure the other errors returned from this function don't trigger retries unnecessarily.

pkg/accelerator-orchestrator/store/group.go
```diff
@@ -0,0 +1,143 @@
...
+
+	pb "github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/accelerator-orchestrator/api/v1alpha1"
+)
+
+// Group represents the in-memory and persistent state of a time-slice group.
+type Group struct {
+	mu    sync.RWMutex
+	id    string
+	nodes []string
+	// activeJob is the job_id holding the lock, empty if none.
```
aishukamal: s/activeJob/lockingJob

pkg/accelerator-orchestrator/store/group_store.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (s *GroupStore) Put(ctx context.Context, group *Group) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	g, ok := s.groups[group.ID()]
+	if ok {
+		group.queue = g.queue
+	}
```
aishukamal: Seems like the intent here is to support an "update". In what scenarios will we need this? If we do want to support update, what other state (e.g. lockingJob, activeJob?) needs to be preserved in addition to queue?

pkg/accelerator-orchestrator/store/group.go
```diff
@@ -0,0 +1,143 @@
...
+// GetWaitingJobQueue returns a reference to the waiter queue for the group.
+func (g *Group) GetWaitingJobQueue() *WaitingJobQueue {
+	g.mu.RLock()
+	defer g.mu.RUnlock()
+	return g.queue
+}
+
+func (g *Group) Delete(ctx context.Context) error {
+	g.mu.Lock()
+	defer g.mu.Unlock()
```
aishukamal: Do we actually need a lock here? Seems like the only object we access is lockStore which has its own locking mechanism and supports concurrent access already?

pkg/accelerator-orchestrator/store/group.go
```diff
@@ -0,0 +1,143 @@
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (g *Group) Lock(ctx context.Context, jobID string) error {
+	g.mu.Lock()
+	defer g.mu.Unlock()
```
aishukamal, line 109: Similar comment here and for Unlock(). I think we only need the lock for the in-memory fields, so after line 125. The access to g.lockStore can be lock-free since it has its own concurrency control mechanism and more importantly, any access to lockStore is a network call which we don't want to do holding this mutex exclusively. 

I would also add a comment in the code on why the locking happens later.

pkg/accelerator-orchestrator/store/job.go
```diff
@@ -0,0 +1,68 @@
...
+	pb "github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/accelerator-orchestrator/api/v1alpha1"
+)
+
+// Job represents the state of a job within a group.
+type Job struct {
+	mu           sync.RWMutex
+	jobID        string
+	groupID      string
+	pods         []string // pod UUIDs
+	contextState map[string]pb.SnapshotAgentJobState_State
```
aishukamal, line 16: What is the key here? Is it nodeID, jobID, podID?

pkg/accelerator-orchestrator/store/job.go
```diff
@@ -0,0 +1,68 @@
...
+	return maps.Clone(j.contextState)
+}
+
+func (j *Job) SetContextState(cs map[string]pb.SnapshotAgentJobState_State) {
+	j.mu.Lock()
+	defer j.mu.Unlock()
+	if len(cs) == 0 {
+		j.contextState = make(map[string]pb.SnapshotAgentJobState_State)
+		return
+	}
```
aishukamal, line 61: maps.Clone() I think handles nil and empty maps just fine, so we can get rid of this.

pkg/accelerator-orchestrator/store/configmap_lockstore.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		if cm.Data == nil {
+			cm.Data = make(map[string]string)
+		}
+
+		current, ok := cm.Data[groupID]
+		if ok && current != "" && current != jobID {
+			return ErrAlreadyLocked
+		}
```
aishukamal: Nit: `if current := cm.Data[groupID]; current != "" && current != jobID { return ErrAlreadyLocked }`. The `ok` check is redundant.

pkg/accelerator-orchestrator/store/job_store.go
```diff
@@ -0,0 +1,90 @@
...
+
+// NewJobStore creates a new JobStore.
+func NewJobStore() *JobStore {
+	return &JobStore{
+		jobs: make(map[string]*Job),
+	}
+}
+
+func makeJobKey(groupID, jobID string) string {
+	return groupID + ":" + jobID
```
aishukamal, line 24: This can cause collisions. E.g. jobID="a:b", groupID="c" vs jobID="a", groupID="b:c". Consider using a struct as the key instead.

E.g.
```
type jobKey struct {
    jobID   string
    groupID string
}

jobs map[jobKey]*Job

func (s *JobStore) Get(ctx context.Context, groupID, jobID string) (*Job, error) {
    key := jobKey{jobID, groupID}
    j, ok := s.jobs[key]
    ...
}
```

pkg/accelerator-orchestrator/store/lock_memstore.go
```diff
@@ -0,0 +1,55 @@
...
+	return nil
+}
+
+// Unlock persistently releases the lock for the group.
+func (s *MemLockStore) Unlock(ctx context.Context, groupID, jobID string) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	current, ok := s.locks[groupID]
+	if !ok || current == "" {
```
aishukamal: Nit: Similar redundant check on `!ok`.

pkg/accelerator-orchestrator/store/lock_memstore.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Lock persistently sets the job_id holding the lock for the group.
+func (s *MemLockStore) Lock(ctx context.Context, groupID, jobID string) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	current, ok := s.locks[groupID]
+	if ok && current != "" && current != jobID {
+		return ErrAlreadyLocked
+	}
```
aishukamal: Nit: redundant check on `ok`. `current` will be "" if the key doesn't exist. Consider `if current := s.locks[groupID]; current != "" && current != jobID {`

pkg/accelerator-orchestrator/store/waiting_job_queue.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// List returns a deep copy of the jobs currently in the queue, maintaining FIFO order.
+func (q *WaitingJobQueue) List() []WaitingJob {
+	q.mu.RLock()
+	defer q.mu.RUnlock()
+
+	if len(q.jobs) == 0 {
+		return nil
+	}
```
aishukamal: Nit: is nil vs empty slice a meaningful distinction for the caller? If not, consider removing.

pkg/accelerator-orchestrator/store/waiting_job_queue.go
```diff
@@ -0,0 +1,106 @@
...
+	res := make([]WaitingJob, len(q.jobs))
+	copy(res, q.jobs)
+	return res
+}
+
+// Reset clears the queue and re-initializes it with the provided jobs, preserving their order.
+// It automatically filters out any duplicates in the input slice (keeping the first occurrence).
+func (q *WaitingJobQueue) Reset(jobs []WaitingJob) {
```
aishukamal: How do you intend to use this? I'm just curious how the caller sets `QueuedSince` on the list of WaitingJobs.

pkg/accelerator-orchestrator/store/waiting_job_queue.go
```diff
@@ -0,0 +1,106 @@
...
+func (q *WaitingJobQueue) Dequeue() (string, bool) {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if len(q.jobs) == 0 {
+		return "", false
+	}
+
+	job := q.jobs[0]
+	q.jobs = q.jobs[1:]
```
aishukamal: Nit: I think this would leak memory, consider zeroing out `q.jobs[0]` before re-slicing.

Alternatively, would a linked-list be a better data structure for the FIFO queue? You could use the standard container/list library for it. My concern with a slice is it is backed by a contiguous array which is ever growing - each job calls into the queue 4 times per step (1 acquire/yield for sample (or train) and 1 acquire/yield for weight transfer) and a job can run 1000s of steps, so we are looking at a pretty large backing array for multiple jobs. Not a blocker for this PR, but worth a thought.

jessicaochen: Good idea, switched to linked list.

pkg/accelerator-orchestrator/store/waiting_job_queue.go
jessicaochen: You are right, I am not sure when this would be used. Removed.

pkg/accelerator-orchestrator/store/waiting_job_queue.go
jessicaochen: Removed.

pkg/accelerator-orchestrator/store/lock_memstore.go
jessicaochen: Cleaned.

pkg/accelerator-orchestrator/store/lock_memstore.go
jessicaochen: Cleaned.

pkg/accelerator-orchestrator/store/group_store.go
jessicaochen: Good catch. I thought about it and I want updates to happen on the group object itself. Therefore I removed this Put & added a GetOrCreate instead.

pkg/accelerator-orchestrator/store/group.go
jessicaochen: Fixed.

pkg/accelerator-orchestrator/store/configmap_lockstore.go
jessicaochen: Yep, AI confirms it is based on a specific code returned that means there was version conflict (& not a general retry)

pkg/accelerator-orchestrator/store/configmap_lockstore_test.go
jessicaochen: Done

pkg/accelerator-orchestrator/store/configmap_lockstore.go
jessicaochen: Moved it into a retry loop :D

pkg/accelerator-orchestrator/store/job_store.go
jessicaochen: I checked the syntatic restrictions for labels. The job IDs you are suggesting are not possible.

pkg/accelerator-orchestrator/store/configmap_lockstore.go
jessicaochen: Removed.

pkg/accelerator-orchestrator/store/job.go
jessicaochen: While there is a key concept (internally) for the jobstore, that key would not belong on this object. 

Afterall, the right key combo depends on the map (and the map's use). JobStore uses groupid-jobid. However, a podStore probably uses podId. etc.

pkg/accelerator-orchestrator/store/group.go
jessicaochen: Nope, Switched it all to methods that themselves handle locking.

pkg/accelerator-orchestrator/store/group.go
jessicaochen: I want the in-memory fields & the lock store state to act as a unit. I do not want readers to get partial state where local state says one thing but we already committed a different value to the actual store. They way to achieve this is with locks.

pkg/accelerator-orchestrator/store/job.go
jessicaochen: It will not panic but will set nil if ti gets nil. I avoid a bunch of nil checks in this class by presuming the map is never nil (only ever empty). Added a comment to be more obvious.

PR #19, opened 2026-06-04
Adding helm charts for the snapshot-agent

Edwinhr716:
## Why is this change needed? 

Adds the helm charts needed to deploy the snapshot daemon. 

Sample command:

```
helm upgrade --install snapshot-agent ./deploy/snapshot-agent   --set image.repository=<image>   --set image.tag=<tag>   --set "tolerations[0].key=nvidia.com/gpu"   --set "tolerations[0].operator=Equal"   --set "tolerations[0].value=present"   --set "tolerations[0].effect=NoSchedule"   --set serviceAccount.create=true   --set serviceAccount.name=snapshot-agent-new-sa
```

## How was this tested?

## Checklist

### DIFF
--- a/deploy/snapshot-agent/templates/daemonset.yaml
+++ b/deploy/snapshot-agent/templates/daemonset.yaml
@@ -1,0 +1,99 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "snapshot-agent.fullname" . }}
+  labels:
+    {{- include "snapshot-agent.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "snapshot-agent.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "snapshot-agent.selectorLabels" . | nindent 8 }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "snapshot-agent.serviceAccountName" . }}
+      hostNetwork: true
+      hostPID: true
+      initContainers:
+        - name: install-cuda-checkpoint
+          image: alpine:latest
+          command: ["sh", "-c", "apk add --no-cache wget && wget -qO /opt/bin/cuda-checkpoint https://raw.githubusercontent.com/NVIDIA/cuda-checkpoint/main/bin/x86_64_Linux/cuda-checkpoint && chmod +x /opt/bin/cuda-checkpoint && echo 'cuda-checkpoint installed'"]
+          volumeMounts:
+            - name: bin-dir
+              mountPath: /opt/bin
+      containers:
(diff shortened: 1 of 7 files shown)


deploy/snapshot-agent/templates/daemonset.yaml
```diff
@@ -0,0 +1,99 @@
...
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "snapshot-agent.serviceAccountName" . }}
+      hostNetwork: true
+      hostPID: true
+      initContainers:
+        - name: install-cuda-checkpoint
+          image: alpine:latest
+          command: ["sh", "-c", "apk add --no-cache wget && wget -qO /opt/bin/cuda-checkpoint https://raw.githubusercontent.com/NVIDIA/cuda-checkpoint/main/bin/x86_64_Linux/cuda-checkpoint && chmod +x /opt/bin/cuda-checkpoint && echo 'cuda-checkpoint installed'"]
```
aishukamal, line 30: Not a blocker for this PR. I wonder if it would be better to fork cuda-checkpoint into the time-slicing repo and build it into the snapshot-agent image at build time rather than pulling the main branch at run time? It would help avoid instabilities/incompatibilities caused by upstream changes to cuda-checkpoint.

deploy/snapshot-agent/values.yaml
```diff
@@ -0,0 +1,56 @@
...
+nodeSelector: {}
+
+tolerations: []
+
+affinity: {}
+
+port: 9001
+
+securityContext:
+  privileged: true
```
aishukamal, line 30: Not a blocker for this PR. We should maybe file a follow-up issue in our backlog to scope the permissions down to exactly what we need rather than broad full host access - for e.g. SYS_PTRACE, SYS_ADMIN, CHECKPOINT_RESTORE capabilities might suffice. This requires some testing, so let's not block this PR on this.

deploy/snapshot-agent/templates/daemonset.yaml
```diff
@@ -0,0 +1,99 @@
...
+          emptyDir: {}
+        {{- if .Values.nvidia.driver.enabled }}
+        - name: nvidia-driver
+          hostPath:
+            path: {{ .Values.nvidia.driver.hostPath }}
+        {{- end }}
+        {{- if .Values.nvidia.devices.enabled }}
+        - name: nvidia-devices
+          hostPath:
+            path: {{ .Values.nvidia.devices.hostPath }}
```
aishukamal, line 87: Should we add `type: Directory` here - https://kubernetes.io/docs/concepts/storage/volumes/#hostpath-volume-types? Based on my read of the documentation, it seems like without this, Kubernetes does no validation that the path exists on the node before mounting and if the path is absent the pod could start silently with a broken mount rather than failing fast.

aishukamal (commented):
/approve

deploy/snapshot-agent/templates/daemonset.yaml
Edwinhr716: Added

deploy/snapshot-agent/templates/daemonset.yaml
Edwinhr716: Not sure what the usual pattern is, but don't think a fork into this repo is the best move. Can we pin it to a commit install that version instead?

PR #20, opened 2026-06-04
Adding core logic for checkpointing

Edwinhr716:
## Why is this change needed?

 Adds the core snapshotting logic using cuda-checkpoint

## How was this tested?

Follow up PRs will include:
- Unit tests will be added in follow up PR
- Storage Bytes and SnapshotDeviceBytes
## Checklist


cmd/snapshot-agent/main.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func main() {
+	port := flag.Int("port", 9001, "The port to listen on")
+	flag.Parse()
+
+	cudaBackend := backends.NewCudaCheckpoint()
+
+	registeredBackends := map[string]backends.Backend{
+		"cuda": cudaBackend,
```
jessicaochen: should the supported backend options be on the API? probably as an enum?

cmd/snapshot-agent/main.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"flag"
+	"log"
+
+	"github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/snapshot-agent/backends"
+	"github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/snapshot-agent/server"
+)
+
+func main() {
+	port := flag.Int("port", 9001, "The port to listen on")
```
jessicaochen, line 28: should this be a const somewhere so the created clients can default to the default port for the agent?

pkg/accelerator-orchestrator/store/configmap_lockstore_test.go
```diff
@@ -50,7 +50,7 @@ func TestConfigMapLockStore_LockAndUnlock(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			ctx := context.Background()
-			client := fake.NewSimpleClientset()
+			client := fake.NewClientset()
```
jessicaochen, line 53: Was it intentional to modify  a test under pkg/accelerator-orchestrator/store? seems irrelevant to your PR scope and logic.

pkg/snapshot-agent/backends/cuda-checkpoint.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, pid := range pids {
+		pidArgs = append(pidArgs, "--pid", pid)
+	}
+
+	if err := c.runSudoCommand(ctx, binaryPath, append([]string{"--action", "lock"}, pidArgs...)...); err != nil {
+		return fmt.Errorf("cuda-checkpoint lock failed: %w", err)
+	}
+	if err := c.runSudoCommand(ctx, binaryPath, append([]string{"--action", "checkpoint"}, pidArgs...)...); err != nil {
+		return fmt.Errorf("cuda-checkpoint checkpoint failed: %w", err)
+	}
```
jessicaochen: (nit) this would be cleaner with a helper function that wraps all the details of the command line.

pkg/snapshot-agent/server/server.go
```diff
@@ -15,52 +20,180 @@ import (
...
+		defaultBackend: defaultBackend,
+	}
 }
 
-// Snapshot implements SnapshotAgentService.Snapshot.
+// Snapshot triggers an asynchronous snapshot of the accelerator context for a job.
 func (s *Server) Snapshot(ctx context.Context, req *pb.SnapshotRequest) (*pb.SnapshotResponse, error) {
-	log.Printf("Snapshot called: JobID=%s, Group=%s", req.GetJobId(), req.GetGroup())
-	return nil, status.Errorf(codes.Unimplemented, "method Snapshot not implemented")
+	log.Printf("Snapshot called: JobID=%s, Group=%s, Backend=%s", req.GetJobId(), req.GetGroup(), req.GetBackend())
```
jessicaochen: (nit) I recommend using a context logger. It will handle making shore all these details stick around as you make further logging statements.

pkg/snapshot-agent/server/server.go
```diff
@@ -15,52 +20,180 @@ import (
...
+			return fmt.Errorf("no pods found for job %s", req.GetJobId())
+		}
+
+		var allPIDs []int
+		var allPIDStrings []string
+		log.Printf("Pods found for job %s: %v", req.GetJobId(), pods)
+		for i := range pods {
+			pod := &pods[i]
+			pids, err := podutils.GetPodPIDs(bgCtx, pod.Name, pod.Namespace)
+			log.Printf("Pod %s has PIDs: %v", pod.Name, pids)
```
jessicaochen: (nit) feels big enough to have a helper fuction

pkg/snapshot-agent/state-machine/state-manager.go
```diff
@@ -0,0 +1,258 @@
...
+	if job.State == pb.JobState_JOB_STATE_TRANSITIONING {
+		return "", status.Errorf(codes.Aborted, "job %s is already transitioning", jobID)
+	}
+
+	// 2. Fault Protection
+	if job.State == pb.JobState_JOB_STATE_FAULTED {
+		return "", status.Errorf(codes.FailedPrecondition, "job %s is in FAULTED state", jobID)
+	}
+
+	opID := fmt.Sprintf("snap-%d", time.Now().UnixNano())
```
jessicaochen: I recommend using an actual UUID instead of time. You avoid tranisent edge cases with it. :)

pkg/snapshot-agent/server/server.go
```diff
@@ -15,52 +20,180 @@ import (
...
+	}
+
+	return resp, nil
 }
 
-// Status implements SnapshotAgentService.Status.
+// Status returns the current state of jobs and accelerators on the node.
 func (s *Server) Status(ctx context.Context, req *pb.StatusRequest) (*pb.StatusResponse, error) {
 	log.Printf("Status called")
 	return nil, status.Errorf(codes.Unimplemented, "method Status not implemented")
```
jessicaochen, line 175: (FYI) no action needed, just letting you know that the orchestrator is depending on this to construct currentsstate for reconciliation.

jessicaochen (changes requested):

pkg/accelerator-orchestrator/store/configmap_lockstore_test.go
Edwinhr716: Yes, otherwise the linter fails

pkg/snapshot-agent/server/server.go
Edwinhr716: ack, decided only to cover the checkpoint logic here. Status and Health endpoints will have their own PRs

pkg/snapshot-agent/server/server.go
Edwinhr716: makes sense, changed to context logger

cmd/snapshot-agent/main.go
Edwinhr716: Makes sense to me, changed to ENUM in the API

pkg/snapshot-agent/state-machine/state-manager.go
Edwinhr716: fixed

cmd/snapshot-agent/main.go
Edwinhr716: Like a const in the python client? Not sure where the go client gets its const from

jessicaochen: I meant put it in a go package somewhere so we can have one source of truth at least on the golang side. Not the end of the world if you cannot since clearly, the yaml and python cannot read golang.

jessicaochen (dismissed):

pkg/snapshot-agent/backends/cuda-checkpoint.go
Edwinhr716: Added

pkg/snapshot-agent/server/server.go
Edwinhr716: done

jessicaochen (approved):

PR #21, opened 2026-06-05
Add reconcile framework to orchestrator

jessicaochen:
Adds the framework as well and initilization.
Determining Goal state and
doing anything will come in subsequent PRs.

## How was this tested?

## Checklist


aishukamal (commented):
Great work on this! Before diving into line-level feedback, I wanted to raise a design question first - what are your thoughts on using the full k8s controller pattern here vs a simpler periodic polling loop that calls the k8s API to list nodes/pods and rebuild group state? Given that group state changes are infrequent and one of our goals is to keep the k8s dependency minimal and as a swappable metadata source (especially if we want to support Slurm or bare-metal in future), I wonder if the informer/workqueue machinery is more than we need. Would love to discuss before you go further.

pkg/accelerator-orchestrator/infrastructure/kubernetes.go
```diff
@@ -0,0 +1,431 @@
...
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+const (
+	NodeLabelPrefix  = "group.timeslice.io/"
+	PodLabelKey      = "timeslice.io/group"
+	JobLabelKey      = "timeslice.io/job-id"
+	JobAnnotationKey = "timeslice.io/job-id"
```
aishukamal: Why do we have both an annotation and label? I thought we were panning to only use labels?

pkg/accelerator-orchestrator/infrastructure/kubernetes.go
```diff
@@ -0,0 +1,431 @@
...
+	pods, err := k.podLister.List(selector)
+	if err != nil {
+		return nil, err
+	}
+	var podInfos []PodInfo
+	for _, pod := range pods {
+		jobID := pod.Annotations[JobAnnotationKey]
+		if jobID == "" {
+			jobID = pod.Labels[JobLabelKey]
+		}
```
aishukamal: Why the fallback? Can we align on one of them - either annotation or label?

pkg/accelerator-orchestrator/infrastructure/kubernetes.go
```diff
@@ -0,0 +1,431 @@
...
+	// 5. Query snapshot agents and update job context states
+	if err := k.updateJobContext(ctx, groupID, groupNodes); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (k *KubernetesOrchestrator) updateGroupNodes(ctx context.Context, groupID string, groupNodes []string) error {
+	logger := klog.FromContext(ctx)
```
aishukamal, line 167: The intent behind using klog here vs slog in other places is understood. However, `klog.FromContext(ctx)` looks up a klog logger stored in the context under klog's own internal key. The context passed here from `processNextWorkItem` carries an slog logger stored under `controller.loggerKey{}` which is a completely different key. `klog.FromContext` won't find it and will fall back to the default klog logger, silently dropping the per-group context (e.g. "group", groupID) that was set up in `processNextWorkItem`. Should we consider replacing `klog.FromContext(ctx)` with `controller.Logger(ctx)` throughout this file to preserve the per-group log context?

pkg/accelerator-orchestrator/server/server.go
aishukamal: Nit: I understand this is existing code. Should we use slog for logging everywhere to be consistent? Same comment applies to main.go. Feel free to update in a follow-up PR, if you agree.

pkg/accelerator-orchestrator/server/server.go
```diff
@@ -48,14 +52,23 @@ func (s *Server) GetGroupStatus(ctx context.Context, req *pb.GetGroupStatusReque
...
+// It also starts the controller in the background.
+func StartServer(ctx context.Context, port int, ctrl *controller.Controller, workers int) error {
 	lis, err := (&net.ListenConfig{}).Listen(ctx, "tcp", fmt.Sprintf(":%d", port))
 	if err != nil {
 		return fmt.Errorf("failed to listen: %w", err)
 	}
 
+	// Start controller in background
+	go func() {
+		log.Printf("Starting controller from server with %d workers...", workers)
```
aishukamal, line 64: Should we use slog for logging everywhere in this file to be consistent since you already use slog in a few other places? Same comment applies to main.go. Feel free to update in a follow-up PR, if you agree.

aishukamal (commented):
LGTM. Just left a few minor comments.

pkg/accelerator-orchestrator/infrastructure/kubernetes.go
jessicaochen: Good catch, that snuck in on me. removign to be just labels.

pkg/accelerator-orchestrator/infrastructure/kubernetes.go
jessicaochen: Good catch, that snuck in on me. removing to be just labels.

pkg/accelerator-orchestrator/infrastructure/kubernetes.go
jessicaochen: I fully agree but since it would touch many files, I will keep the switch to a consistent logging in a different PR.

pkg/accelerator-orchestrator/server/server.go
jessicaochen: I fully agree but since it would touch many files, I will keep the switch to a consistent logging in a different PR.

PR #23, opened 2026-06-05
Add cli and update deployment

jessicaochen:
Add needed service account to authorize orchestrator to read k8s state.

Add cli with command to talk to a deployed orchestrator

Valdidated by deploying the orchestator and using cli to talk to it as documented in readme

## How was this tested?

## Checklist


cmd/rlts/cmd/orchestrator.go
```diff
@@ -0,0 +1,176 @@
...
+			fmt.Printf("  State Timestamp: %s\n", resp.Group.StateTimestamp.AsTime())
+		}
+		fmt.Printf("  Locking Job: %s\n", resp.Group.LockingJob)
+		fmt.Printf("  Active Job: %s\n", resp.Group.ActiveJob)
+		fmt.Printf("  Waiter Queue Depth: %d\n", resp.Group.WaiterQueueDepth)
+
+		if len(resp.AgentJobStates) > 0 {
+			fmt.Println("  Agent Job States:")
+			for _, state := range resp.AgentJobStates {
+				fmt.Printf("    - Agent: %s, State: %s\n", state.Agent, state.JobState)
```
aishukamal, line 117: Nit: I think this prints the integer value in string format by default. You probably need to call `.String()`.

cmd/rlts/cmd/orchestrator.go
```diff
@@ -0,0 +1,176 @@
...
+
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+
+		resp, err := client.GetGroupStatus(ctx, &pb.GetGroupStatusRequest{GroupId: groupID})
+		if err != nil {
+			return fmt.Errorf("failed to get status: %w", err)
+		}
+
+		fmt.Printf("Group: %s\n", resp.Group.GroupId)
```
aishukamal, line 105: Is `Group` guaranteed to be non-nil?

deploy/acceleratororchestrator/templates/clusterrole.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: {{ include "acceleratororchestrator.fullname" . }}
+  labels:
+    {{- include "acceleratororchestrator.labels" . | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "nodes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["configmaps"]
```
aishukamal: Do we actually need full ConfigMap read/write access across the entire cluster? The orchestrator only needs access to a single ConfigMap in one namespace. Should this be a Role + RoleBinding scoped to that namespace instead of a ClusterRole + ClusterRoleBinding?

cmd/rlts/cmd/orchestrator.go
jessicaochen: Technically, no.

Practically, yes. If there is no group or there is any errors fetching it, it would have exited in the error check earlier.

I'll add the check but it does mean that the server has a bug if it manages to get to this state.

cmd/rlts/cmd/orchestrator.go
jessicaochen: This actually works as is. the %s will autocall the .String() for me.

deploy/acceleratororchestrator/templates/clusterrole.yaml
jessicaochen: Took a few more files but did the restriction & updated the docs.

PR #25, opened 2026-06-08
Adding python client for snapshot agent

Edwinhr716:
## What does this PR do?
Adds python client to interact with the snapshot agent. 

Blocked by #20, #26, #27

## How was this tested?

## Checklist


pkg/client/python/timeslice/snapshot_agent/client.py
```diff
@@ -0,0 +1,103 @@
...
+        """Close the gRPC channel."""
+        self.channel.close()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+
+    def snapshot(self, job_id: str, group: str, backend: str = ""):
```
aishukamal: Is `group` actually required for snapshot agent given that it's local to a node and we're already making an assumption that a node belongs to a single group only?

@jessicaochen can keep me honest here.

pkg/client/python/timeslice/snapshot_agent/client.py
```diff
@@ -0,0 +1,103 @@
...
+            response = self.stub.Snapshot(request)
+            logger.info(f"Snapshot Response: operation_id={response.operation_id}")
+            return response
+        except grpc.RpcError as e:
+            logger.error(f"gRPC error calling Snapshot: {e.code()} - {e.details()}")
+        except Exception as e:
+            logger.error(f"Unexpected error in Snapshot: {e}")
+        return None
+
+    def get_operation(self, operation_id: str):
```
aishukamal: Should we also provide a helper function, say, `snapshot_and_wait()` that polls `get_operation` until complete? Avoids requiring each caller to reimplement polling.

pkg/client/python/pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[project]
+name = "timeslice"
+version = "0.1.0"
+description = "Timeslice library including Snapshot Agent client"
+readme = "README.md"
+requires-python = ">=3.8"
+dependencies = [
+    "grpcio>=1.64.1",
+    "requests>=2.32.3",
```
aishukamal: Why is this needed? We don't use HTTP anywhere, do we?

pkg/client/python/timeslice/snapshot_agent/client.py
```diff
@@ -0,0 +1,103 @@
...
+            )
+            logger.info(f"Calling Snapshot with job_id={job_id}, group={group}...")
+            response = self.stub.Snapshot(request)
+            logger.info(f"Snapshot Response: operation_id={response.operation_id}")
+            return response
+        except grpc.RpcError as e:
+            logger.error(f"gRPC error calling Snapshot: {e.code()} - {e.details()}")
+        except Exception as e:
+            logger.error(f"Unexpected error in Snapshot: {e}")
+        return None
```
aishukamal: One thing that's bothering me is `snapshot` and `restore` APIs return None on error whereas `get_operation()`returns the response object on both success and error. So the caller now has two completely different patterns for error handling. Should we consolidate to a single contract - either have all 3 APIs raise exception on error or (the more graceful and cleaner option) always return a response object that contains both the response and any error?

pkg/client/python/timeslice/snapshot_agent/client.py
Edwinhr716: At least in the API implementation we don't use group for anything at the moment, but it is a field in the request API which is why it was added here.

aishukamal: If it is not used I would remove it. Just adds unnecessary friction for the user. Feel free to clean up in a follow-up CR if you'd like.

Edwinhr716: Will do it in a follow up PR

pkg/client/python/timeslice/snapshot_agent/client.py
Edwinhr716: Added helper function

pkg/client/python/pyproject.toml
Edwinhr716: removed http

pkg/client/python/timeslice/snapshot_agent/client.py
Edwinhr716: Refactored to follow AIP standards instead. Used this package as reference https://github.com/googleapis/google-cloud-python/tree/main/packages/google-cloud-gkerecommender/google/cloud/gkerecommender_v1

BogdanAtWork (approved):
Looks good to me. To things to consider as follow ups:

1. Clean up the API surface with removing `group` for now (defaulting to `""` for instance) and adding potential async APIs that will `await`-ing snapshot and restore operations more naturally. 
2. Add test coverage to ensure that the API surface for the client doesn't change accidentally.

pkg/client/python/timeslice/snapshot_agent/client.py
```diff
@@ -0,0 +1,321 @@
...
+        while True:
+            response = self.get_operation(operation_id)
+            if response.status in [
+                "OPERATION_STATUS_COMPLETE",
+                "OPERATION_STATUS_FAILED",
+            ]:
+                return response
+            time.sleep(poll_interval_sec)
+
+    def snapshot_and_wait(
```
aishukamal, line 301: Might want to support a configurable timeout for both `snapshot_and_wait` and `restore_and_wait`. Also default the timeout to a non-infinite (but large enough) value. Can be done in a follow-up PR.

pkg/client/python/timeslice/snapshot_agent/types.py
```diff
@@ -0,0 +1,59 @@
...
+    """Response message for Health Check RPC."""
+
+    status: str
+
+
+@dataclass(frozen=True)
+class GetOperationResponse:
+    """Response message for GetOperation RPC."""
+
+    status: str
```
aishukamal, line 30: There are a bunch of status/state fields here that are currently stored as strings converted from proto enums via `.Name()`. I wonder what the standard pattern here is - is it to use IntEnum types that mirror the proto enum values or use the converted string values? The advantage of the former is it avoids magic string comparisons like response.status == "OPERATION_STATUS_COMPLETE" and allows direct comparison with proto integer values.

Not a blocker for this PR if you choose to go with IntEnum.

pkg/client/python/timeslice/snapshot_agent/client.py
jessicaochen: Sorry for missing this flag. My suggestion would be to make groupid optional in client/gprc but keep the concept of group id in agent.

That way direct users of the agent do not need to specify groupid & we will not be backed into a corner when we have more than one group per node.

PR #26, opened 2026-06-09
Health endpoint implementation

Edwinhr716:
## Why is this change needed?
Implements the health endpoint following grpc standard defined in https://github.com/grpc/grpc/blob/master/doc/health-checking.md

## How was this tested?

## Checklist

### DIFF
--- a/pkg/snapshot-agent/backends/checkpoint.go
+++ b/pkg/snapshot-agent/backends/checkpoint.go
@@ -21,6 +21,7 @@ type Backend interface {
 	// Restore triggers a restoration of the accelerator context for a job.
 	Restore(ctx context.Context, pids []string) error
 
-	// Discover checks if the backend is healthy and returns any error found.
+	// Discover checks if the backend is healthy by initializing the backend
+	// and the discovery provider.
 	Discover(ctx context.Context) error
 }
--- a/pkg/snapshot-agent/backends/checkpoint.go
+++ b/pkg/snapshot-agent/backends/checkpoint.go
@@ -20,4 +20,7 @@ type Backend interface {
 
 	// Restore triggers a restoration of the accelerator context for a job.
 	Restore(ctx context.Context, pids []string) error
+
+	// Discover checks if the backend is healthy and returns any error found.
+	Discover(ctx context.Context) error
 }
--- a/pkg/snapshot-agent/backends/checkpoint.go
+++ b/pkg/snapshot-agent/backends/checkpoint.go
@@ -21,7 +21,7 @@ type Backend interface {
 	// Restore triggers a restoration of the accelerator context for a job.
 	Restore(ctx context.Context, pids []string) error
 
-	// Discover checks if the backend is healthy by initializing the backend
+	// Health checks if the backend is healthy by initializing the backend
 	// and the discovery provider.
-	Discover(ctx context.Context) error
+	Health(ctx context.Context) error
 }
--- a/pkg/snapshot-agent/backends/checkpoint.go
+++ b/pkg/snapshot-agent/backends/checkpoint.go
@@ -21,7 +21,7 @@ type Backend interface {
 	// Restore triggers a restoration of the accelerator context for a job.
 	Restore(ctx context.Context, pids []string) error
 
-	// Health checks if the backend is healthy by initializing the backend
-	// and the discovery provider (if it uses one).
-	Health(ctx context.Context) error
+	// HealthCheck checks if the backend is healthy by initializing the backend
+	// and the discovery provider.
+	HealthCheck(ctx context.Context) error
 }
--- a/pkg/snapshot-agent/backends/checkpoint.go
+++ b/pkg/snapshot-agent/backends/checkpoint.go
@@ -22,6 +22,6 @@ type Backend interface {
 	Restore(ctx context.Context, pids []string) error
 
 	// Health checks if the backend is healthy by initializing the backend
-	// and the discovery provider.
+	// and the discovery provider (if it uses one).
 	Health(ctx context.Context) error
 }
--- a/pkg/snapshot-agent/backends/cuda-checkpoint.go
+++ b/pkg/snapshot-agent/backends/cuda-checkpoint.go
@@ -102,7 +102,8 @@ func (c *CudaCheckpoint) restorePIDs(ctx context.Context, pids []string) error {
 	return nil
 }
 
-// Discover checks if the cuda-checkpoint backend is healthy.
+// Discover checks if the cuda-checkpoint backend is healthy by initializing the backend
+// and the discovery provider.
 func (c *CudaCheckpoint) Discover(ctx context.Context) error {
 	// 1. Check if cuda-checkpoint executable is available
 	binaryPath := c.getCudaCheckpointPath()
--- a/pkg/snapshot-agent/backends/cuda-checkpoint.go
+++ b/pkg/snapshot-agent/backends/cuda-checkpoint.go
@@ -103,6 +103,7 @@ func (c *CudaCheckpoint) restorePIDs(ctx context.Context, pids []string) error {
 
 // Discover checks if the cuda-checkpoint backend is healthy.
 func (c *CudaCheckpoint) Discover(ctx context.Context) error {
+	logger := klog.FromContext(ctx)
 	// 1. Check if cuda-checkpoint executable is available
 	binaryPath := c.getCudaCheckpointPath()
 	if _, err := exec.LookPath(binaryPath); err != nil {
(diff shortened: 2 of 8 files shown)


pkg/snapshot-agent/server/server.go
```diff
@@ -177,8 +178,19 @@ func (s *Server) Status(ctx context.Context, req *pb.StatusRequest) (*pb.StatusR
...
+	slog.Info("Health called", "backend", req.GetBackend())
+
+	backendType := s.getBackendType(req.GetBackend())
+
+	backend, ok := s.backendMap[backendType]
+	if !ok {
+		return nil, status.Errorf(codes.NotFound, "backend %s not found", backendType)
+	}
+
+	if err := backend.Discover(ctx); err != nil {
```
jessicaochen: Not fully sure of the use case but do we want to return anyways with a Healthy: false?

pkg/snapshot-agent/server/server_test.go
```diff
@@ -135,8 +135,28 @@ func TestServer_Health(t *testing.T) {
...
 	if err != nil {
-		t.Errorf("Expected success, got error: %v", err)
+		t.Fatalf("Expected success for default backend, got error: %v", err)
+	}
+	if !resp.Healthy {
+		t.Errorf("Expected healthy=true for default backend")
+	}
+
+	// Test explicit noop backend (it maps to default in this test setup because Cuda is not registered)
+	// Actually, getBackendType handles it.
```
jessicaochen: (nit) can this comment be less like a running thought train and just say the final conclusion?

pkg/snapshot-agent/backends/checkpoint.go
```diff
@@ -20,4 +20,8 @@ type Backend interface {
 
 	// Restore triggers a restoration of the accelerator context for a job.
 	Restore(ctx context.Context, pids []string) error
+
+	// Discover checks if the backend is healthy by initializing the backend
+	// and the discovery provider.
+	Discover(ctx context.Context) error
```
jessicaochen: (nit) Not sure if you want to be this specific about how health is determined at the interface level. Consider the backends we are thinking of supporting. Do we expect them all to have discovery providers?

maybe something closer to "Health" might provide a more future proof interface level

jessicaochen (changes requested):

pkg/snapshot-agent/server/server.go
Edwinhr716: Added healthy : false

pkg/snapshot-agent/server/server_test.go
Edwinhr716: removed

pkg/snapshot-agent/backends/checkpoint.go
Edwinhr716: Yeah makes sense to not expose that in the interface. renamed to health

pkg/snapshot-agent/server/server.go
```diff
@@ -177,8 +178,19 @@ func (s *Server) Status(ctx context.Context, req *pb.StatusRequest) (*pb.StatusR
...
+
+	backendType := s.getBackendType(req.GetBackend())
+
+	backend, ok := s.backendMap[backendType]
+	if !ok {
+		return nil, status.Errorf(codes.NotFound, "backend %s not found", backendType)
+	}
+
+	if err := backend.Health(ctx); err != nil {
+		return nil, status.Errorf(codes.Internal, "health check failed: %v", err)
```
jessicaochen: I could be missing something but I do not see the path where we will have a Healthy: false

Your response to my previous comment seem to say you were planning to have that path when HealthCheck failed

pkg/snapshot-agent/backends/checkpoint.go
```diff
@@ -20,4 +20,8 @@ type Backend interface {
 
 	// Restore triggers a restoration of the accelerator context for a job.
 	Restore(ctx context.Context, pids []string) error
+
+	// Health checks if the backend is healthy by initializing the backend
+	// and the discovery provider (if it uses one).
+	Health(ctx context.Context) error
```
jessicaochen: (nit) 

Health would make sense for signature Health(...) bool, error

XXX(...) error is a signature for an action that can complete successfully or not. If you want to keep this signature, it is best named after an action that can be failed. eg. HealthCheck, EnsureHealth, EnsureReady...

jessicaochen: (yes, I know I said Health instead of Discovery in the last comment. While the implications/scope still holds, I did not think through the form of the name for the signature)

jessicaochen (changes requested):

pkg/snapshot-agent/server/server.go
Edwinhr716: oops forgot to push the commit. Added now

pkg/snapshot-agent/backends/checkpoint.go
Edwinhr716: Makes sense, changed it to HealthCheck. Opted to keep the signature since bool would just match whether or not an error was hit when initializing, so conveys the same message.

pkg/snapshot-agent/server/server.go
```diff
@@ -187,8 +187,8 @@ func (s *Server) Health(ctx context.Context, req *pb.HealthRequest) (*pb.HealthR
 		return nil, status.Errorf(codes.NotFound, "backend %s not found", backendType)
 	}
 
-	if err := backend.Health(ctx); err != nil {
-		return nil, status.Errorf(codes.Internal, "health check failed: %v", err)
+	if err := backend.HealthCheck(ctx); err != nil {
+		return &pb.HealthResponse{Healthy: false}, status.Errorf(codes.Internal, "health check failed: %v", err)
```
jessicaochen: Are you sure you want to return a failure status on the gRPC? The server was able to check health, it was just not healthy.

If you want to record details on why it is not healthy, perhaps consider logging and/or having more details in HealthResponse proto.

jessicaochen (dismissed):

pkg/snapshot-agent/server/server.go
Edwinhr716: Makes sense, fixed it to only return false, not an error

pkg/snapshot-agent/backends/cuda-checkpoint.go
```diff
@@ -99,3 +101,35 @@ func (c *CudaCheckpoint) restorePIDs(ctx context.Context, pids []string) error {
...
+		return fmt.Errorf("cuda-checkpoint executable not found: %w", err)
+	}
+
+	// 2. Initialize NVML
+	if ret := nvml.Init(); ret != nvml.SUCCESS {
+		return fmt.Errorf("failed to initialize NVML: %v", nvml.ErrorString(ret))
+	}
+	defer func() {
+		if ret := nvml.Shutdown(); ret != nvml.SUCCESS {
+			slog.Error("Failed to shutdown NVML", "error", nvml.ErrorString(ret))
```
aishukamal, line 120: There is use of klog in some places and slog in others. Let's converge on using slog everywhere (which is what we are doing on the orchestrator side as well). Can be done in a follow-up PR.

pkg/snapshot-agent/server/server_test.go
```diff
@@ -133,10 +135,36 @@ func TestServer_Health(t *testing.T) {
...
+	}
+	if resp.Status != grpc_health_v1.HealthCheckResponse_SERVING {
+		t.Errorf("Expected status=SERVING for empty service, got %v", resp.Status)
+	}
+
+	resp, err = client.Check(ctx, &grpc_health_v1.HealthCheckRequest{Service: string(backends.BackendNoop)})
+	if err != nil {
+		t.Fatalf("Expected success for noop backend, got error: %v", err)
+	}
+	if resp.Status != grpc_health_v1.HealthCheckResponse_SERVING {
```
aishukamal, line 171: Can we also test the NOT_SERVING case?

pkg/snapshot-agent/backends/cuda-checkpoint.go
```diff
@@ -99,3 +101,35 @@ func (c *CudaCheckpoint) restorePIDs(ctx context.Context, pids []string) error {
...
+// and the discovery provider.
+func (c *CudaCheckpoint) HealthCheck(ctx context.Context) error {
+	// 1. Check if cuda-checkpoint executable is available
+	binaryPath := c.getCudaCheckpointPath()
+	if _, err := exec.LookPath(binaryPath); err != nil {
+		return fmt.Errorf("cuda-checkpoint executable not found: %w", err)
+	}
+
+	// 2. Initialize NVML
+	if ret := nvml.Init(); ret != nvml.SUCCESS {
```
aishukamal, line 115: I don't think initializing nvlm is cheap and it is probably an overkill to re-initialize on every call. Actually, all 3 checks (cuda executable available, nvlm initialized and devices attached) are more or less static and rarely change. So I would do them once (or until they succeed) and cache these results. For now, maybe the health check serves simply as a liveness check but in the future we can probably add a ttl to the cached results and refresh every once in a while, if necessary. WDYT?

pkg/snapshot-agent/backends/cuda-checkpoint.go
Edwinhr716: yep, we agreed with @jessicaochen to use slog in new PRs, and migrate the codebase to slog after June 15th

pkg/snapshot-agent/backends/cuda-checkpoint.go
Edwinhr716: Makes sense to cache the results, although we run the risk of returning healthy when it is currently unhealthy. I'll create an issue for this implementation and we can discuss it further.

pkg/snapshot-agent/server/server_test.go
Edwinhr716: Addded

PR #27, opened 2026-06-09
Implementing status endpoint

Edwinhr716:
## Why is this change needed?

Implements the status endpoint of snapshot-agent

## How was this tested?

## Checklist

### DIFF
--- a/pkg/snapshot-agent/server/server_test.go
+++ b/pkg/snapshot-agent/server/server_test.go
@@ -117,9 +117,33 @@ func TestServer_Status(t *testing.T) {
 	defer conn.Close()
 	client := pb.NewSnapshotAgentServiceClient(conn)
 
-	_, err = client.Status(ctx, &pb.StatusRequest{})
+	jobID := "test-job-status"
+	_, err = client.Snapshot(ctx, &pb.SnapshotRequest{
+		JobId:   jobID,
+		Group:   "test-group",
+		Backend: pb.Backend_BACKEND_UNSPECIFIED,
+	})
+	if err != nil {
+		t.Fatalf("Snapshot failed: %v", err)
+	}
+
+	resp, err := client.Status(ctx, &pb.StatusRequest{})
 	if err != nil {
 		t.Errorf("Expected success, got error: %v", err)
+	}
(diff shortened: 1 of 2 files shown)


pkg/snapshot-agent/server/server_test.go
```diff
@@ -118,8 +118,8 @@ func TestServer_Status(t *testing.T) {
 	client := pb.NewSnapshotAgentServiceClient(conn)
 
 	_, err = client.Status(ctx, &pb.StatusRequest{})
```
jessicaochen: should we check what is returned has what we expect? ( in empty scenario but also if there are jobs with status in your state store)

jessicaochen (changes requested):

pkg/snapshot-agent/server/server_test.go
Edwinhr716: Added

Edwinhr716: Added both cases

pkg/snapshot-agent/server/server_test.go
```diff
@@ -117,10 +117,43 @@ func TestServer_Status(t *testing.T) {
...
+	resp, err := client.Status(ctx, &pb.StatusRequest{})
 	if err != nil {
 		t.Errorf("Expected success, got error: %v", err)
 	}
+
+	if len(resp.JobStatuses) != 0 {
+		t.Errorf("Expected no jobs, got %d", len(resp.JobStatuses))
+	}
+
+	jobID := "test-job-status"
```
jessicaochen, line 129: (nit) this probably should be in golang testcase format instead of serial inline test cases. That ensures there are no ordering shenanigans that slip in.

jessicaochen (approved):

pkg/snapshot-agent/server/server_test.go
Edwinhr716: I'll add more comprehensive unit tests in follow up PRs. I'll change to testcase format there, for now it isn't clear what the test case would look like since the two cases we have require different setups.

PR #30, opened 2026-06-09
Orchestrator Get and List Methods

jessicaochen:
Also added a missing field into the proto.

## How was this tested?

## Checklist


pkg/accelerator-orchestrator/server/server.go
```diff
@@ -42,18 +60,75 @@ func (s *Server) Yield(ctx context.Context, req *pb.YieldRequest) (*pb.YieldResp
...
+		return nil, status.Errorf(codes.Unavailable, "group %s state is unknown", req.GetGroupId())
+	}
+
+	groupStatus := &pb.GroupStatus{
+		GroupId:          group.ID(),
+		GroupState:       state,
+		StateTimestamp:   timestamppb.New(stateTime),
+		LockingJob:       group.LockingJob(),
+		ActiveJob:        group.ActiveJob(),
+		WaiterQueueDepth: int64(group.GetWaitingJobQueue().Len()),
```
aishukamal: Two minor points to consider with this approach:
1. We are returning an inconsistent state because we make multiple calls to group. So group state could be LOCKED but lockingJob could be empty if a yield happens between the two calls.
2. We are acquiring and releasing the Group mutex 4 times for each GetGroupStatus() call.

Consider adding a `Snapshot()` or equivalent method to Group that returns all status fields under a single lock and avoids inconsistent reads when building the GroupStatus response.

jessicaochen: This is a good point, will work on getting a atomic reads for this method

jessicaochen: Added a Clone to get an atomic copy. Personally chose a deepclone instead of a dedicated snapshot struct since the prior is more maintainable.

pkg/accelerator-orchestrator/store/group.go
```diff
@@ -140,3 +140,31 @@ func (g *Group) Unlock(ctx context.Context, jobID string) error {
...
+
+	return &Group{
+		id:             g.id,
+		nodes:          nodes,
+		lockingJob:     g.lockingJob,
+		activeJob:      g.activeJob,
+		state:          g.state,
+		stateTimestamp: g.stateTimestamp,
+		queue:          clonedQueue,
+		lockStore:      g.lockStore,
```
aishukamal: `lockStore` is a reference to the actual lockStore object, isn't it? If the caller decides to call `Lock()` or `Unlock()` on it, it will actually mutate state. Should we nil-ify it before returning?

jessicaochen: switched from clone to snapshot struct to avoid this type of confusion (ie. trade more to maintain for less error prone)

PR #31, opened 2026-06-10
Add orchestrator desired state determination

jessicaochen:
Also organized the desrired state into spec for easier mental reasoning between what we want (ie. spec) and what is currently the state.

Note that logging is just extending what is currently used, it will all flip to the standard with the other PR.

## How was this tested?

## Checklist

### DIFF
--- a/pkg/accelerator-orchestrator/store/group.go
+++ b/pkg/accelerator-orchestrator/store/group.go
@@ -81,6 +81,9 @@ func (g *Group) SetNodes(nodes []string) {
 
 // Spec returns the GroupSpec for the group.
 func (g *Group) Spec() *GroupSpec {
+	// No lock is safe here because the pointer is not
+	// modifiable after the Group is created. Fields
+	// on the spec itself are controlled by the internal mutex.
 	return g.spec
 }
 
@@ -102,13 +105,7 @@ func (g *Group) SetState(state pb.GroupStatus_State) {
 
 // Delete deletes the group by releasing its lock if it is currently held.
 func (g *Group) Delete(ctx context.Context) error {
-	spec := g.Spec()
-	lj := spec.LockingJob()
-	if lj == "" {
-		return nil
-	}
-
-	return spec.Unlock(ctx, lj)
+	return g.spec.Delete(ctx)
 }
 
 // Methods on GroupSpec
@@ -134,18 +131,14 @@ func (s *GroupSpec) GetWaitingJobQueue() *WaitingJobQueue {
 	return s.queue
 }
 
-// Lock acquires the lock for the given job ID.
-func (s *GroupSpec) Lock(ctx context.Context, jobID string) error {
+// Delete cleans up the spec information.
+func (s *GroupSpec) Delete(ctx context.Context) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	return s.lock(ctx, jobID)
-}
-
-// Unlock releases the lock for the given job ID.
-func (s *GroupSpec) Unlock(ctx context.Context, jobID string) error {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	return s.unlock(ctx, jobID)
+	if s.lockingJob == "" {
+		return nil
+	}
+	return s.unlock(ctx, s.lockingJob)
 }
 
 // lock assumes s.mu is held.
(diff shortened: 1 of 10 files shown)


pkg/accelerator-orchestrator/store/group.go
```diff
@@ -99,44 +100,114 @@ func (g *Group) SetState(state pb.GroupStatus_State) {
...
 func (g *Group) Delete(ctx context.Context) error {
-	lj := g.LockingJob()
+	spec := g.Spec()
+	lj := spec.LockingJob()
 	if lj == "" {
 		return nil
 	}
 
-	return g.Unlock(ctx, lj)
+	return spec.Unlock(ctx, lj)
```
aishukamal: `spec.Unlock()` doesn't clear `activeJob`. We probably want to clear it for the deletion case?

pkg/accelerator-orchestrator/store/group.go
```diff
@@ -65,22 +79,9 @@ func (g *Group) SetNodes(nodes []string) {
...
-	return g.activeJob
-}
-
-func (g *Group) SetActiveJob(jobID string) {
-	g.mu.Lock()
-	defer g.mu.Unlock()
-	g.activeJob = jobID
+// Spec returns the GroupSpec for the group.
+func (g *Group) Spec() *GroupSpec {
+	return g.spec
```
aishukamal, line 111: This doesn't need the lock because spec has its own lock? It's not obvious when we need to acquire the `Group` lock - maybe worth adding a comment.

pkg/accelerator-orchestrator/store/group.go
```diff
@@ -99,44 +100,114 @@ func (g *Group) SetState(state pb.GroupStatus_State) {
...
+		return err
+	}
+
+	nextJobID, ok := s.queue.Peek()
+	if !ok {
+		return nil
+	}
+
+	if err := s.lock(ctx, nextJobID); err != nil {
+		return fmt.Errorf("yield succeeded in unlocking but failed to lock next job %s: %w", nextJobID, err)
```
aishukamal: What is the expected action from the caller here? Is `Yield` idempotent so the caller can simply retry in this case?

pkg/accelerator-orchestrator/store/group.go
jessicaochen: Hmmm..... I think my choice of naming perhaps has led to confusion. Yield & RequestLock are what move the state machine and the path on which customer calls with come. Unlock/Lock are cleanup/test hooks. Let me think if there is a way to name this better.....

pkg/accelerator-orchestrator/store/group.go
jessicaochen: Tried to make it more obvious by breaking out a status object and removing at the group level.

pkg/accelerator-orchestrator/store/group.go
jessicaochen: interesting point. I personally prefer not to block the yielder on the next job promoting. Moved job promotion logic to the reconciliation loop.

pkg/accelerator-orchestrator/store/group.go
aishukamal: My question was more on the `Delete()` API. Who is the `Delete()` API intended for and what is it supposed to do other than unlock the group? My point was that `Delete()` internally calls `unlock()` which has a specific comment stating "we do not clear the active job. This is because we actually want to leave the context on the machines until there is a new job that wants to lock the group.". But if we truly want to "delete" we should be clearing everything including the active job, no?

pkg/accelerator-orchestrator/store/group.go
aishukamal: Just so I understand the diff correctly:
1. User calls Yield(), jobA gets queued
2. User calls Acquire(), jobB gets queued

In the old version, when the Yield() call is dequeued, it would release the lock for jobA but also peek into the queue and if there is a waiter (in this case jobB), it would acquire the lock for jobB. In the next reconcile loop, the Acquire() call for jobB would be processed but since the previous Yield() already processed it for jobB, it would moot.

In the new version, when Yield() is dequeued, it would release the lock for jobA and return. Then when reconciliation dequeues the Acquire() call for jobB, it would acquire the lock for jobB.

Am I understanding this correctly?

pkg/accelerator-orchestrator/store/group.go
jessicaochen: The scenario where delete happens is when a group existed then all the nodes are gone so now the group no longer exists. No need to clean up context because there are no nodes to the group. 

We just need to cleanup the other state to avoid leaking etc. so clients, the lock (since you cannot exactly lock a empty group of nodes and if the group comes back, we do not want it in a pre-locked state) etc.

pkg/accelerator-orchestrator/store/group.go
jessicaochen: Yes, the promotion of jobB to lock holder moved from the yield call to the reconcile call.

PR #39, opened 2026-06-14
Adding user guide on how to use snapshot-agent

Edwinhr716:
## What does this PR do?

Adds user guide on how to deploy the snapshot-agent, and how to use the python client

## How was this tested?

## Checklist

### DIFF
--- a/guides/multitenant-managed-fft-timeslicing/README.md
+++ b/guides/multitenant-managed-fft-timeslicing/README.md
@@ -24,10 +24,7 @@ By using the Snapshot Agent, you can implement time-slicing without modifying th
 The Snapshot Agent must be deployed as a DaemonSet to ensure it is available on every node.
 
 ### Using Helm
-```bash
-cd ../../deploy/
-helm install snapshot-agent ./snapshot-agent
-```
+Follow the instrcutions in [helm-snapshot-agent.md](../../deploy/snapshot-agent/README.md) to deploy the Snapshot Agent.
 
 ### Key Configuration (`values.yaml`)
 - `port`: The gRPC port (default: `9001`).
@@ -66,9 +63,15 @@ env:
 The `timeslice.SnapshotAgentClient` library provides a high-level Python API for interacting with the Snapshot Agent.
 
 ### Installation
+From the local repository:
 ```bash
 cd ../../pkg/client/python
 pip install .
+```
+
+Or install from the remote repository:
+```bash
+pip install "git+https://github.com/llm-d-incubation/llm-d-rl-time-slicing.git#subdirectory=pkg/client/python"
 ```
 
 ### Choosing a Backend
@@ -99,7 +102,7 @@ with SnapshotAgentClient(endpoint) as client:
     print("Snapshotting...")
     result = client.snapshot_and_wait(job_id, group)
     if result.status == "OPERATION_STATUS_COMPLETE":
-        print(f"Snapshot success! Used {result.storage_bytes} bytes")
+        print(f"Snapshot success! Completed in {result.elapsed_ms} ms")
 
     # ... wait for your turn (orchestrated externally) ...
 
@@ -107,7 +110,7 @@ with SnapshotAgentClient(endpoint) as client:
     print("Restoring...")
     result = client.restore_and_wait(job_id, group)
     if result.status == "OPERATION_STATUS_COMPLETE":
-        print("Restore success!")
+        print("Restore success! Completed in {result.elapsed_ms} ms")
 ```
 
 ### Advanced Usage
--- a/guides/multitenant-managed-fft-timeslicing/README.md
+++ b/guides/multitenant-managed-fft-timeslicing/README.md
@@ -1,6 +1,6 @@
-# Snapshotting GPU workloads with Snapshot Agent
+# GPU Time-Slicing for Multi-Tenant Managed RL Services
 
-This guide describes how to use the **Snapshot Agent** to enable snapshotting for LLM workloads. This allows multiple tenants to share GPU resources efficiently by snapshotting and restoring GPU state (memory and registers).
+This guide is for developers building a managed RL service who want to enable multiple tenants to share GPU nodes by time-slicing their RL workloads. This applies to any GPU-bound RL component — trainer, sampler, or both, each running as a separate pod. The service is responsible for deciding when to preempt a running job and when to resume it. The Snapshot Agent provides the GPU checkpoint/restore primitive the service calls at those boundaries. From the tenant's perspective, their RL job simply pauses and resumes transparently
 
 ## Overview
 
@@ -14,7 +14,7 @@ By using the Snapshot Agent, you can implement time-slicing without modifying th
 ## Architecture
 
 1.  **Snapshot Agent (DaemonSet):** Runs on every node with GPUs. It has privileged access to the GPU devices and host paths required for snapshotting.
-2.  **Application Pod (Workload):** Your LLM workload (e.g., vLLM, PyTorch) running in a pod.
+2.  **Tenant's RL Workload:**: A sampler or trainer running on a pod.
 3.  **Snapshot Agent Client:** A Python library used by the application pod to communicate with the local Snapshot Agent via gRPC (port 9001).
 
 ---
@@ -38,7 +38,7 @@ See [helm-snapshot-agent.md](../../deploy/snapshot-agent/README.md) for more det
 
 ---
 
-## 2. Integrating with Your Workload
+## 2. Integrating with Tenant's RL Workload
 
 To enable a pod for time-slicing, you need to add specific labels and environment variables to your Deployment.
 
@@ -70,6 +70,18 @@ The `timeslice.SnapshotAgentClient` library provides a high-level Python API for
 cd ../../pkg/client/python
 pip install .
 ```
+
+### Choosing a Backend
+The `backend` parameter controls how GPU state is checkpointed. It is an optional argument in the following `SnapshotAgentClient` methods:
+
+*   `snapshot(job_id, group, backend=...)`
+*   `restore(job_id, group, backend=...)`
+*   `snapshot_and_wait(job_id, group, backend=...)`
+*   `restore_and_wait(job_id, group, backend=...)`
+
+Available backends:
+*   **BACKEND_CUDA (default):** Full process-level GPU checkpoint via `cuda-checkpoint`. Suitable for any GPU-bound RL component (trainer, sampler) running full model weights.
+*   Additional backends for lighter-weight workloads (e.g. LoRA adapters) are planned for a future release.
 
 ### Basic Workflow
 The most common usage is to trigger a snapshot at the end of a "slice" and a restore at the beginning of the next one.
--- a/guides/snapshot-agent/README.md
+++ b/guides/snapshot-agent/README.md
@@ -88,7 +88,7 @@ The `backend` parameter controls how GPU state is checkpointed. It is an optiona
 *   `restore_and_wait(job_id, backend=...)`
 
 Available backends:
-*   **BACKEND_CUDA (default):** Full process-level GPU checkpoint via `cuda-checkpoint`. Suitable for any GPU-bound component (trainer, sampler, LLM engine) running full model weights.
+*   **BACKEND_CUDA (default):** Full process-level GPU checkpoint via `cuda-checkpoint`. It offloads the complete accelerator state (VRAM) to host memory. Suitable for any GPU-bound component (trainer, sampler, LLM engine) running full model weights.
 *   Additional backends for lighter-weight workloads (e.g. LoRA adapters) are planned for a future release.
 
 ### Basic Workflow
--- a/guides/snapshot-agent/README.md
+++ b/guides/snapshot-agent/README.md
@@ -1,6 +1,11 @@
-# GPU Time-Slicing for Multi-Tenant Managed RL Services
+# Snapshot Agent
 
-This guide is for developers building a managed RL service who want to enable multiple tenants to share GPU nodes by time-slicing their RL workloads. This applies to any GPU-bound RL component — trainer, sampler, or both, each running as a separate pod. The service is responsible for deciding when to preempt a running job and when to resume it. The Snapshot Agent provides the GPU checkpoint/restore primitive the service calls at those boundaries. From the tenant's perspective, their RL job simply pauses and resumes transparently
+The Snapshot Agent provides GPU checkpoint/restore primitives to enable efficient resource sharing for GPU-bound workloads. By allowing processes to save and reload their entire GPU state, it enables scenarios where multiple high-memory workloads can share the same physical GPU hardware.
+
+## Use Cases
+
+1.  **Multi-tenant GPU Time-slicing:** Enable multiple tenants to share GPU nodes by time-slicing their workloads (e.g., RL trainers and samplers). The Snapshot Agent provides the primitives to preempt and resume these jobs transparently.
+2.  **Multi-model vLLM Server Proxy:** Serve multiple models (e.g., same family such as Gemma4 or Qwen3) on the same GPU by taking turns for inference requests. A proxy can snapshot the GPU state of one model and restore another to handle incoming requests for different models without requiring them to fit in GPU memory simultaneously.
 
 ## Overview
 
@@ -9,12 +14,12 @@ The Snapshot Agent is a gRPC service that runs as a DaemonSet on each GPU node. 
 - **Restore:** Reload a previously saved state to the GPU.
 - **Status:** Monitor the current state of jobs and accelerators.
 
-By using the Snapshot Agent, you can implement time-slicing without modifying the core LLM engine code, simply by wrapping your execution loop with the `timeslice.SnapshotAgentClient` client library.
+By using the Snapshot Agent, you can implement context switching without modifying the core engine code, simply by wrapping your execution loop with the `timeslice.SnapshotAgentClient` client library.
 
 ## Architecture
 
 1.  **Snapshot Agent (DaemonSet):** Runs on every node with GPUs. It has privileged access to the GPU devices and host paths required for snapshotting.
-2.  **Tenant's RL Workload:**: A sampler or trainer running on a pod.
+2.  **Workload:** A GPU-bound process (e.g., a sampler, trainer, or vLLM instance) running in a pod.
 3.  **Snapshot Agent Client:** A Python library used by the application pod to communicate with the local Snapshot Agent via gRPC (port 9001).
 
 ---
@@ -24,7 +29,7 @@ By using the Snapshot Agent, you can implement time-slicing without modifying th
 The Snapshot Agent must be deployed as a DaemonSet to ensure it is available on every node.
 
 ### Using Helm
-Follow the instrcutions in [helm-snapshot-agent.md](../../deploy/snapshot-agent/README.md) to deploy the Snapshot Agent.
+Follow the instructions in [helm-snapshot-agent.md](../../deploy/snapshot-agent/README.md) to deploy the Snapshot Agent.
 
 ### Key Configuration (`values.yaml`)
 - `port`: The gRPC port (default: `9001`).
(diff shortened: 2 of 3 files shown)


guides/snapshot-workloads/README.md
```diff
@@ -0,0 +1,140 @@
+# Snapshotting GPU workloads with Snapshot Agent
```
aishukamal: I would suggest a title like "GPU Time-Slicing for Multi-Tenant Managed RL Services" or something along those lines.

guides/snapshot-workloads/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+This guide describes how to use the **Snapshot Agent** to enable snapshotting for LLM workloads. This allows multiple tenants to share GPU resources efficiently by snapshotting and restoring GPU state (memory and registers).
```
aishukamal: Maybe provide some context on the target use case here? I was thinking something like:

"This guide is for developers building a managed RL service who want to enable multiple tenants to share GPU nodes by time-slicing their RL workloads. This applies to any GPU-bound RL component — trainer, sampler, or both, each running as a separate pod. The service is responsible for deciding when to preempt a running job and when to resume it. The Snapshot Agent provides the GPU checkpoint/restore primitive the service calls at those boundaries. From the tenant's perspective, their RL job simply pauses and resumes transparently."

guides/snapshot-workloads/README.md
```diff
@@ -0,0 +1,140 @@
...
+- **Snapshot:** Save the current state of a GPU job.
+- **Restore:** Reload a previously saved state to the GPU.
+- **Status:** Monitor the current state of jobs and accelerators.
+
+By using the Snapshot Agent, you can implement time-slicing without modifying the core LLM engine code, simply by wrapping your execution loop with the `timeslice.SnapshotAgentClient` client library.
+
+## Architecture
+
+1.  **Snapshot Agent (DaemonSet):** Runs on every node with GPUs. It has privileged access to the GPU devices and host paths required for snapshotting.
+2.  **Application Pod (Workload):** Your LLM workload (e.g., vLLM, PyTorch) running in a pod.
```
aishukamal: "A tenant's RL workload (e.g., vLLM, PyTorch) running in a pod" instead of "Your LLM workload" since the target user is the managed RL service developer?

guides/multitenant-managed-fft-timeslicing/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+---
+
+## 1. Deploying the Snapshot Agent
+
+The Snapshot Agent must be deployed as a DaemonSet to ensure it is available on every node.
+
+### Using Helm
+```bash
+cd ../../deploy/
```
aishukamal: Nit: should we just use the repo root relative path? E.g. `helm install snapshot-agent ./deploy/snapshot-agent`

guides/snapshot-workloads/README.md
```diff
@@ -0,0 +1,140 @@
...
+### Key Configuration (`values.yaml`)
+- `port`: The gRPC port (default: `9001`).
+- `securityContext.privileged`: Must be `true` to access GPU registers.
+- `nvidia.driver.hostPath`: Path to NVIDIA driver binaries on the host (e.g., `/home/kubernetes/bin/nvidia`).
+
+See [helm-snapshot-agent.md](../../deploy/snapshot-agent/README.md) for more details on the Helm chart.
+
+---
+
+## 2. Integrating with Your Workload
```
aishukamal: Nit: "Integrating with Your Service"?

guides/snapshot-agent/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+## 2. Integrating with Your Workload
+
+To enable a pod for time-slicing, you need to add specific labels and environment variables to your Deployment.
+
+### Required Labels
+The Snapshot Agent identifies pods that should be managed via labels:
+- `timeslice.io/job-id: "<unique-job-id>"`: A unique identifier for the job.
+
+### Environment Variables
```
aishukamal, line 51: Should we add a "Choosing a Backend" section here? Something like:

## Choosing a Backend

The `backend` parameter controls how GPU state is checkpointed:
- `BACKEND_CUDA` (default): Full process-level GPU checkpoint via cuda-checkpoint.
  Suitable for any GPU-bound RL component (trainer, sampler) running full model weights.
- Additional backends for lighter-weight workloads (e.g. LoRA adapters) are
  planned for a future release.

aishukamal (commented):
The current guide is too generic and doesn't convey a specific use case or target audience. Should we focus this well-lit path on enabling time-slicing for multi-tenant managed RL services?

(All my other comments assume this guide would cater to this well-lit path).

aishukamal:
Adding @droot to provide feedback on the well-lit path.

guides/snapshot-workloads/README.md
Edwinhr716: I hesitated to name it that since we want to target more general use case outside of time-slicing correct? That's why I went with a more generic title. Happy to discuss further though

droot:
> Adding @droot to provide feedback on the well-lit path.

Thanks. Will take a look at it shortly.

/cc @ShubyM

guides/snapshot-agent/README.md
```diff
@@ -0,0 +1,140 @@
+# Snapshotting GPU workloads with Snapshot Agent
+
+This guide describes how to use the **Snapshot Agent** to enable snapshotting for LLM workloads. This allows multiple tenants to share GPU resources efficiently by snapshotting and restoring GPU state (memory and registers).
+
+## Overview
+
+The Snapshot Agent is a gRPC service that runs as a DaemonSet on each GPU node. It provides a local interface for application pods to:
+- **Snapshot:** Save the current state of a GPU job.
+- **Restore:** Reload a previously saved state to the GPU.
+- **Status:** Monitor the current state of jobs and accelerators.
```
droot, line 15: It's important to talk about persistence here. Typically, state is persisted in storage such as disk/gcs etc, but in our case, state if saved in host memory and restored from host memory. We can mention other storage backend are on the roadmap. Eventually, we will end up supporting some sort of tired storage (cpu --> disk --> remote storage such as gcs etc.)

We should tie it to the motivation and use-cases which is primarily offloading the accelerator state (VRAM) to host memory (DRAM) in this case.

guides/multitenant-managed-fft-timeslicing/README.md
```diff
@@ -0,0 +1,140 @@
...
+This guide describes how to use the **Snapshot Agent** to enable snapshotting for LLM workloads. This allows multiple tenants to share GPU resources efficiently by snapshotting and restoring GPU state (memory and registers).
+
+## Overview
+
+The Snapshot Agent is a gRPC service that runs as a DaemonSet on each GPU node. It provides a local interface for application pods to:
+- **Snapshot:** Save the current state of a GPU job.
+- **Restore:** Reload a previously saved state to the GPU.
+- **Status:** Monitor the current state of jobs and accelerators.
+
+By using the Snapshot Agent, you can implement time-slicing without modifying the core LLM engine code, simply by wrapping your execution loop with the `timeslice.SnapshotAgentClient` client library.
```
droot: If we have timeslicing docs somewhere in the public domain, we should add a link to those. Also, I would recommend, not to assume any linkage with timeslicing when writing this guide.

guides/multitenant-managed-fft-timeslicing/README.md
Edwinhr716: Thoughts on adding `$REPO_ROOT`? That's the pattern that llm-d uses now for their user guides https://github.com/llm-d/llm-d/blob/main/guides/wide-ep-lws/README.md?plain=1#L101

guides/snapshot-workloads/README.md
Edwinhr716: What about: integrate with the tenant's RL Workload? Since this covers what needs to be added to the pod

guides/multitenant-managed-fft-timeslicing/README.md
```diff
@@ -0,0 +1,140 @@
...
+---
+
+## 1. Deploying the Snapshot Agent
+
+The Snapshot Agent must be deployed as a DaemonSet to ensure it is available on every node.
+
+### Using Helm
+```bash
+cd ../../deploy/
+helm install snapshot-agent ./snapshot-agent
```
droot: helm feels like an overkill for this because there isn't much customization. consider providing static manifests that one can deploy using `kubectl -f <>`

guides/snapshot-agent/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+The Snapshot Agent must be deployed as a DaemonSet to ensure it is available on every node.
+
+### Using Helm
+```bash
+cd ../../deploy/
+helm install snapshot-agent ./snapshot-agent
+```
+
+### Key Configuration (`values.yaml`)
```
droot, line 34: I am assuming atleast on GKE cluster, one doesn't have change these defaults.

guides/multitenant-managed-fft-timeslicing/README.md
droot: I don't see any steps for building the docker image. Are we hosting pre-built images ?

guides/snapshot-workloads/README.md
Edwinhr716: Added

guides/snapshot-workloads/README.md
Edwinhr716: Changed

guides/snapshot-agent/README.md
Edwinhr716: Added under Using python client

guides/snapshot-workloads/README.md
Edwinhr716: Discussed offline. We are going forward with the multi-tenant managed RL services. For other use cases we will either add new user guides or extend this one in the future.

guides/multitenant-managed-fft-timeslicing/README.md
```diff
@@ -0,0 +1,140 @@
...
+- `securityContext.privileged`: Must be `true` to access GPU registers.
+- `nvidia.driver.hostPath`: Path to NVIDIA driver binaries on the host (e.g., `/home/kubernetes/bin/nvidia`).
+
+See [helm-snapshot-agent.md](../../deploy/snapshot-agent/README.md) for more details on the Helm chart.
+
+---
+
+## 2. Integrating with Your Workload
+
+To enable a pod for time-slicing, you need to add specific labels and environment variables to your Deployment.
```
droot: I am not sure the scope of this guide but it will be good to have a guide that demonstrate the usage snapshot-agent in context of a scenario that requires offloading the context etc.

It's okay to have a synthetic scenario. One somewhat useful scenario I can think of:

vllm-swap that lets you run inference on different models (think multiple vllm instances on the backend or wrapped) on the same hardware.

Based on some simple heuristics where it buffers requests and then serve them while taking turns for different models. We can assume same family of models to reuse cuda graph etc.

guides/snapshot-agent/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+---
+
+## 2. Integrating with Your Workload
+
+To enable a pod for time-slicing, you need to add specific labels and environment variables to your Deployment.
+
+### Required Labels
+The Snapshot Agent identifies pods that should be managed via labels:
+- `timeslice.io/job-id: "<unique-job-id>"`: A unique identifier for the job.
```
droot, line 49: we should use something like: `timeslicing.k8s.io` or `timeslicing.llmd.io` or something. llmd might have a convention around the labels/annotations, so lets use that.

We don't know timeslice.io or own this domain... so better to avoid it.

guides/snapshot-agent/README.md
```diff
@@ -0,0 +1,140 @@
...
+    valueFrom:
+      fieldRef:
+        fieldPath: status.hostIP
+  - name: AGENT_ENDPOINT
+    value: "$(NODE_IP):9001"
+```
+
+---
+
+## 3. Using the Python Client
```
droot, line 66: Developer guide for Snapshot-Agent has two CUJs:

1. Development CUJ is mostly where dev will be using a nvidia GPU VM directly
2. Once happy with dev, deploy on kubernetes

I would like the guide to also cover the development CUJ because kubernetes comes with lots of overhead and rarely one develops on kubernetes.

guides/snapshot-agent/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+The `timeslice.SnapshotAgentClient` library provides a high-level Python API for interacting with the Snapshot Agent.
+
+### Installation
+```bash
+cd ../../pkg/client/python
+pip install .
```
droot, line 74: This requires cloning the repo, would be nice if we can host it in the repo or something. I have seen some instructions where people do `pip install https+` paths.

guides/snapshot-agent/README.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+### Basic Workflow
+The most common usage is to trigger a snapshot at the end of a "slice" and a restore at the beginning of the next one.
+
+```python
+from timeslice.snapshot_agent import SnapshotAgentClient
+
+# AGENT_ENDPOINT is typically $(NODE_IP):9001
+endpoint = os.getenv("AGENT_ENDPOINT", "localhost:9001")
```
droot, line 102: it will be nice if one doesn't have to specify these for defaults.

guides/multitenant-managed-fft-timeslicing/README.md
```diff
@@ -0,0 +1,140 @@
...
+### Basic Workflow
+The most common usage is to trigger a snapshot at the end of a "slice" and a restore at the beginning of the next one.
+
+```python
+from timeslice.snapshot_agent import SnapshotAgentClient
+
+# AGENT_ENDPOINT is typically $(NODE_IP):9001
+endpoint = os.getenv("AGENT_ENDPOINT", "localhost:9001")
+job_id = "test-job"
+group = "default"
```
droot: it will be good to explain `group`. may be it is explained somewhere.. I missed it.

guides/multitenant-managed-fft-timeslicing/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+with SnapshotAgentClient(endpoint) as client:
+    # 1. Trigger Snapshot and wait for completion
+    print("Snapshotting...")
+    result = client.snapshot_and_wait(job_id, group)
```
droot: python has idiomatic primitives for async programming... we should ask antigravity to if we can make it more idiomatic.

guides/snapshot-workloads/README.md
```diff
@@ -0,0 +1,140 @@
...
+endpoint = os.getenv("AGENT_ENDPOINT", "localhost:9001")
+job_id = "test-job"
+group = "default"
+
+with SnapshotAgentClient(endpoint) as client:
+    # 1. Trigger Snapshot and wait for completion
+    print("Snapshotting...")
+    result = client.snapshot_and_wait(job_id, group)
+    if result.status == "OPERATION_STATUS_COMPLETE":
+        print(f"Snapshot success! Used {result.storage_bytes} bytes")
```
droot: more than storage bytes, the time taken is useful metric to log.

guides/multitenant-managed-fft-timeslicing/README.md
```diff
@@ -0,0 +1,140 @@
...
+    if result.status == "OPERATION_STATUS_COMPLETE":
+        print(f"Snapshot success! Used {result.storage_bytes} bytes")
+
+    # ... wait for your turn (orchestrated externally) ...
+
+    # 2. Trigger Restore and wait for completion
+    print("Restoring...")
+    result = client.restore_and_wait(job_id, group)
```
droot: same comment.

guides/multitenant-managed-fft-timeslicing/README.md
Edwinhr716: Steps are on the "how to deploy with helm user guide". Will make that explicit

guides/snapshot-agent/README.md
Edwinhr716: Correct, these are set to the correct values by default (at least for GKE)

guides/multitenant-managed-fft-timeslicing/README.md
Edwinhr716: Part of the backlog https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/41

guides/snapshot-agent/README.md
Edwinhr716: As in, have the python client have those values by default?

guides/multitenant-managed-fft-timeslicing/README.md
Edwinhr716: At least in OSS, helm or kustomize is the standard, not static manifests. Also, providing a static manifest would also mean that we need to add GKE specific nodeSelector, and we ideally want to keep it cloud agnostic.

guides/snapshot-agent/README.md
Edwinhr716: Agreed, created https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/43

guides/snapshot-agent/README.md
Edwinhr716: done

guides/snapshot-workloads/README.md
Edwinhr716: done

guides/multitenant-managed-fft-timeslicing/README.md
Edwinhr716: done

guides/multitenant-managed-fft-timeslicing/README.md
droot: sounds good.

guides/snapshot-agent/README.md
droot: yeah. but not a huge deal.

guides/snapshot-agent/README.md
Edwinhr716: Isn't the development CUJ covered by using the Python client?

guides/multitenant-managed-fft-timeslicing/README.md
Edwinhr716: removed linkage to make it more generic

guides/multitenant-managed-fft-timeslicing/README.md
Edwinhr716: Removed it as an option

guides/multitenant-managed-fft-timeslicing/README.md
Edwinhr716: We don't have any docs AFAIK. 

Restructured the guide to remove linkage to timeslicing

guides/multitenant-managed-fft-timeslicing/README.md
Edwinhr716: Added synthetic scenario as well

guides/multitenant-managed-fft-timeslicing/README.md
Edwinhr716: Removing group as a parameter in https://github.com/llm-d-incubation/llm-d-rl-time-slicing/pull/50

guides/snapshot-agent/README.md
Edwinhr716: Added this information under CUDA backend

guides/snapshot-agent/README.md
droot: Development CUJ basically means developing with running snapshot-agent on a non-k8s environment.

Think of the user journey. Say an AI researcher who has a RL Loop implemented using verl framework, they typically test the loop on nvidia host directly in a non-k8s environment. The guide here assume AI researcher will integrate snapshot-agent by running their RL loop on k8s environment directly.

We don't have to block this guide for that.

droot (approved):
LGTM. I have a comment about development CUJ but that doesn't have to block this guide. We can address that in a separate PR (probably need a separate guide for that).

Will defer to @aishukamal to merge it.

PR #44, opened 2026-06-16
Orchestrator server aquire & yield methods

jessicaochen:
## How was this tested?

## Checklist

### DIFF
--- a/pkg/accelerator-orchestrator/server/server.go
+++ b/pkg/accelerator-orchestrator/server/server.go
@@ -125,7 +125,38 @@ func (s *Server) Yield(ctx context.Context, req *pb.YieldRequest) (*pb.YieldResp
 	ctx = logging.WithJobID(ctx, req.GetJobId())
 	ctx = logging.WithGroupID(ctx, req.GetGroupId())
 	slog.InfoContext(ctx, "Yield called")
-	return nil, status.Errorf(codes.Unimplemented, "method Yield not implemented")
+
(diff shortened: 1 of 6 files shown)


pkg/accelerator-orchestrator/server/server.go
```diff
@@ -52,7 +55,68 @@ func (s *Server) Acquire(ctx context.Context, req *pb.AcquireRequest) (*pb.Acqui
...
+			if faulted {
+				return nil, status.Errorf(codes.Unavailable, "group %s is faulted", groupID)
+			}
+
+			// Check if loaded job is ours
+			if group.Status().LoadedJob() == jobID {
+				slog.InfoContext(ctx, "Acquire succeeded, job loaded")
+				return &pb.AcquireResponse{
+					Success:         true,
+					ContextRestored: true, // Default to true, as we don't have enough info to determine if it was zero-overhead
```
aishukamal, line 99: You could technically check if `LoadedJob` was already `jobID` when the `Acquire` call started to determine if ContextRestored should be true or false, no?

jessicaochen: That is actually insufficient since for RL job init, we need to support a job locking with no pods deployed (ie. no context restored). I can circle back on this later (after bare bones MVP) to fill this out by looking at job state *before* we call for AquireLock OR just remove the field. Pretty sure whether context is restored is just a observability thing & does not change acual caller behavior.

PR #45, opened 2026-06-16
Orc Node Reconciliation Logic

jessicaochen:
## How was this tested?

## Checklist

### DIFF
--- a/pkg/accelerator-orchestrator/controller/controller_internal_test.go
+++ b/pkg/accelerator-orchestrator/controller/controller_internal_test.go
@@ -9,8 +9,6 @@ import (
 	agentpb "github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/snapshot-agent/api/v1alpha1"
 )
 
-
-
 func TestController_WaitForOperation(t *testing.T) {
 	nodeName := "node-1"
 	opID := "op-123"
--- a/pkg/accelerator-orchestrator/controller/controller_internal_test.go
+++ b/pkg/accelerator-orchestrator/controller/controller_internal_test.go
@@ -1,0 +1,139 @@
+package controller
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/accelerator-orchestrator/store"
+	agentpb "github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/snapshot-agent/api/v1alpha1"
+)
+
+type mockAgentStoreForInternal struct {
+	store.SnapshotAgentStore
+	getStatusFunc func(ctx context.Context, nodeName string) (*agentpb.StatusResponse, error)
+	snapshotFunc  func(ctx context.Context, nodeName, jobID, groupID string) (*agentpb.SnapshotResponse, error)
+	operationFunc func(ctx context.Context, nodeName, operationID string) (*agentpb.GetOperationResponse, error)
+}
+
+func (m *mockAgentStoreForInternal) GetStatus(ctx context.Context, nodeName string) (*agentpb.StatusResponse, error) {
+	if m.getStatusFunc != nil {
+		return m.getStatusFunc(ctx, nodeName)
+	}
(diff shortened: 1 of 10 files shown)


pkg/accelerator-orchestrator/store/snapshot_agent_store.go
```diff
@@ -15,6 +15,9 @@ import (
 type SnapshotAgentStore interface {
 	GetStatus(ctx context.Context, nodeName string) (*agentpb.StatusResponse, error)
 	CloseClient(nodeName string) error
+	Snapshot(ctx context.Context, nodeName, jobID, groupID string) (*agentpb.SnapshotResponse, error)
+	GetOperation(ctx context.Context, nodeName, operationID string) (*agentpb.GetOperationResponse, error)
+	Restore(ctx context.Context, nodeName, jobID, groupID string) (*agentpb.RestoreResponse, error)
```
aishukamal, line 20: Snapshot agent client provides a handy wrapper that automatically calls Snapshot/Restore RPCs and polls for success - https://github.com/llm-d-incubation/llm-d-rl-time-slicing/blob/main/pkg/client/python/timeslice/snapshot_agent/client.py#L301. You probably only need the SnapshotAndWait and RestoreAndWait APIs here.

pkg/accelerator-orchestrator/controller/controller_internal_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/accelerator-orchestrator/store"
+	agentpb "github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/snapshot-agent/api/v1alpha1"
+)
+
+type mockAgentStoreForInternal struct {
```
aishukamal: Can we reuse `mockSnapshotAgentStore` from controller_internal_test.go here?

pkg/accelerator-orchestrator/controller/controller_internal_test.go
```diff
@@ -0,0 +1,166 @@
...
+		return m.operationFunc(ctx, nodeName, operationID)
+	}
+	return &agentpb.GetOperationResponse{}, nil
+}
+
+func TestController_WaitForOperation(t *testing.T) {
+	nodeName := "node-1"
+	opID := "op-123"
+
+	pendingCalls := 0
```
aishukamal: Nit: this is shared state between tests. Consider resetting in t.Run.

pkg/accelerator-orchestrator/store/snapshot_agent_store.go
jessicaochen: It does not make sense to call python client from go code. If we want to make an official go client (as opposed to the generated one) for agent, I am happy to switch this over to use that in the future.  @Edwinhr716

jessicaochen: Though, I do not see a need for an official go client unless we anticipate customers of this project to be calling in go. My impression is most RL is in python.

pkg/accelerator-orchestrator/controller/controller_internal_test.go
jessicaochen: Good catch. I ended up restructuring it to avoid the shared state.

pkg/accelerator-orchestrator/controller/controller_internal_test.go
jessicaochen: I assume you meant to share the mock between the two controller tests. I ended up pushing the mock into its own file to avoid circular dependency. Now both use one implemenation.

pkg/accelerator-orchestrator/store/snapshot_agent_store.go
aishukamal: Oh, I was not implying calling the python client from go code. I assumed we would have a go client similar to the python client that you can hook into. But you have a point - a local go client interface should suffice for now since we don't see the need for an external client at this point.

PR #49, opened 2026-06-17
Update project README with project details

aishukamal:
## What does this PR do?

Updates the project README with project details.

## How was this tested?

## Checklist


README.md
```diff
@@ -1,77 +1,58 @@
...
+  
+  ## The Solution: Platform-Level Sharing
+  **llm-d-rl-time-slicing** moves the utilization fix from the application layer to the platform layer. Multiple independent RL jobs cooperatively share the same accelerator hardware, swapping during each job's natural blocking phases (generation, training, weight sync) rather than holding the accelerator idle. 
+  
+  **Your training loop stays exactly the same — no algorithmic rewrites required.**
+  
+  ## How It Works
+  We introduce **collaborative, application-aware time-slicing**. Using a lightweight client library that pairs seamlessly with your existing training and inference frameworks, the system delivers two core capabilities:
+  * **Intelligent Scheduling:** Dynamically coordinates accelerator access across concurrent jobs based on their execution phases.
+  * **Ultra-Fast Context Switching:** Performs blazing-fast, transparent state checkpointing and restoration under the hood.
```
jessicaochen: (nit) Can we stick with just "fast"? the blazing and ultra bit sound a bit like exaggeration. Alternatively, you can give a sense of magnitude relative to something else. Eg. fast compared to killing and restarting the service using the GPU?

README.md
```diff
@@ -1,77 +1,58 @@
...
-# Install pre-commit hooks
-pre-commit install
-
-# Build
-make build
+## Architecture
 
-# Run tests
-make test
+![Architecture](https://github.com/llm-d-incubation/llm-d-rl-time-slicing/blob/main/docs/diagrams/time-slicing-architecture-diagram.png?raw=true)
```
jessicaochen, line 25: I will be honest, I think this graph will actually be confusing for folks first familiarizing with this concept. I can mentally sort it out but only because I am already familiar with what the project/concepts are about.

Perhaps it might be cleaner if we broke this up into different things we are communicating:
(A) What component/code is calling what (and for what purpose). Either organized by where the code runs OR in a code stack heirarchy.
(B) The sequencing that happens

More specific points of confusion from the graph:
* We actually expect the RL job or RL framework to deploy the actors while under lock. There is no links indicating this in the graph. 
* The RL loop driver being *in* the trainer and sampler nodepools is odd since that loop driver really can run anywhere & probably there is only one copy. not a copy in each nodepool.
* The trainer nodepool has the agent interfacting with 2/3 of the active jobs but with no label as to the interaction so it is very unclear what that interaction means (ie. does it mean these two are loaded, is one getting a restore while the other one is getting snapshotted at the same time?)
* There are quite a few more. Do let me know when and if you want more.

jessicaochen (changes requested):

README.md
aishukamal: Ack, I'll tone it down to just "fast" for now. Once we have the other planned snapshot backends in place we can probably quantify it and say sub-second or something along those lines.

README.md
aishukamal: Thanks for the detailed feedback!

The intent of this diagram is to show the system's components and their high-level relationships - what lives where and what talks to what. The specific sequencing of events (which phase acquires a lock when, actor deployment under lock, etc.) belongs in a sequence diagram, not the architecture diagram. I'm happy to include a sequence diagram if you'd like.

That said, I've updated the diagram to address the specific points you raised:

> We actually expect the RL job or RL framework to deploy the actors while under lock

The acquire/yield arrow from the RL loop driver to the Orchestrator is intentionally general - any GPU-touching phase (init, train, generate, etc.) must go through it. The diagram shows the gate exists; the specific details of what happens under that lock is protocol-level detail that fits better in a sequence diagram.

> The RL loop driver being in the trainer and sampler nodepools is odd since that loop driver really can run anywhere

Agreed this was confusing. I've pulled the loop driver out of the trainer and sampler boxes so it's clearly a standalone component. It can technically run on a GPU node too, but separating it out better communicates that it doesn't need to run on a GPU node.

> probably there is only one copy. not a copy in each nodepool

Hence the dotted line on the other copy to indicate that it's the same component shown in two places instead of one to avoid having a messy mesh of arrows. With the updated diagram, this should be more clear.

> agent interfacting with 2/3 of the active jobs

I've updated the Snapshot Agent to show uniform connections to all job processes, not just some. 

> but with no label as to the interaction so it is very unclear what that interaction means

The interactions aren't labeled because the snapshot/restore mechanics are internal to the agent - there's no public API boundary there to reference. The agent box itself is labeled "Snapshot Agent - snapshot / restore" which gives enough high-level context IMO.

> There are quite a few more. Do let me know when and if you want more.

If you have a specific alternative layout in mind, I'm happy to iterate on that. Otherwise I'll push this updated version and we can go from there.

jessicaochen: Much improved.

Here are some additional feedback on the current version:
* It does not make sense that "actors places on pods" is the label between the distributed engine runtime and accelerator operator. If that arrow is figuring the where and when of pod placement. It should say aquire/yield to the orchestrator & "which pool? (I am sure you have a better wording)" to the optimizer. If it is about the pods getting deployed then the arrow should point to the Job A sampler, job B smpler, job A trainer .....
* "pods placed on nodes" going into distributed runtime engine is confusing as well . I think it comes from the graph trying to cover both use of a distributed runtime engine and not using one (bare pods). Perhaps stick with one for simplicity? or perhaps those both need to be in one big box as the pod provisioning system?
* (nit) I am not sure what the A active ... boxes provide in terms of new information to the user. I think they do presume that if a box is somewhere it is running.

jessicaochen (changes requested):

README.md
aishukamal: > "actors placed on pods" / arrow between runtime engine and Accelerator Operator

You're right, that interaction is misleading. I've combined the Workload Scheduler and Distributed Runtime Engine into a single "Workload Provisioning" box and removed the arrows between them and the Accelerator Operator since they don't interact. I'm also open to removing this box altogether since it's not a new component. The only reason I have it is to convey where the time-slicing components fit relative to the user's existing infrastructure.

> "pods placed on nodes"

Addressed by the change above - the combined box just lists the components without implying a specific flow between them.

>  think it comes from the graph trying to cover both use of a distributed runtime engine and not using one (bare pods).

Even with Ray, the user still interacts with both - they submit a RayCluster manifest to the Kubernetes scheduler first, then submit RL jobs to the Ray cluster. So these are not entirely separate paths, they're sequential steps in the same flow. But either way, none of that is specific to time-slicing, so the combined box keeps things simpler.

> A active / B active bars:

The idea was to illustrate the _time-slicing_ invariant, not just that jobs exist on the node. Without them, the diagram shows co-location but not the mutual exclusion (i.e. only one job holds the accelerator at a time on each node) that's the core concept of the system.

jessicaochen: I think this is iterating to better places, I do have some feedback on the current iteration:
* I would either leave the client boxes out OR have the arrow of the actions come out of that box.
* When referring to other pieces, I would use the name in the graph eg. "context switching snapshot agent" instead of "context-switcher daemon"
* The idea was to illustrate the time-slicing invariant ... mutual exclusion. It could just be me but the presence of the boxes already shows co-location & the text saying "one job holds accelerator at a time" says the mutual exclusion. That means for me, the active boxes are not saying anything not already said by the rest of the chart. That said, there is no harm if you feel others will glean better understanding with those active boxes. An alternative could be showing the accelerator memory & machine memory with boxes for the jobs that are loaded to highlight that only one is loaded in acclerator when both are in machine memory. 

That said, I will make changes as approved so we can have something filled in on the front page.

jessicaochen (approved):

PR #52, opened 2026-06-18
Making group field optional in Snapshot Agent 

Edwinhr716:
## What does this PR do?
Makes group field optional in Snapshot and Restore API calls in the Snapshot-Agent API. Replaces #50 

## Why is this change needed?
Group isn't needed for all use cases of the Snapshot-Agent, but it is needed for others (e.g the Orchestrator). Because of that, we need to keep it. 

## How was this tested?

## Checklist

### DIFF
--- a/pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
+++ b/pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
@@ -25,6 +25,8 @@ enum Backend {
 }
 
 message SnapshotRequest {
+  // job_id is the unique identifier for the job. If group is specified,
+  // job_id should be unique within that group.
   string job_id = 1;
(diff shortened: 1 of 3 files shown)


pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
```diff
@@ -36,6 +37,7 @@ message SnapshotResponse {
 
 message RestoreRequest {
   string job_id = 1;
+  // group is an optional identifier for a set of related jobs.
```
jessicaochen, line 44: (nit) I think we should also clarify that job_id is expected to be unique per group (if specified) as opposed to be unique globally.

jessicaochen (dismissed):

jessicaochen (approved):

PR #56, opened 2026-06-23
Add agent port to orchestrator routing

jessicaochen:
## What does this PR do?

Adds the agent port to the routing logic

## How was this tested?

- [x] Manual testing performed - it works in cluster in e2e (a different change)

## Checklist

### DIFF
--- a/pkg/accelerator-orchestrator/store/snapshot_agent_store.go
+++ b/pkg/accelerator-orchestrator/store/snapshot_agent_store.go
@@ -3,6 +3,7 @@ package store
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/accelerator-orchestrator/store/snapshot_agent_store.go
```diff
@@ -198,7 +201,10 @@ func (s *GRPCSnapshotAgentStore) CloseClient(nodeName string) error {
 }
 
 func (s *GRPCSnapshotAgentStore) resolveNodeAddress(nodeName string) string {
-	// TODO: Implement actual node name to address translation once we know the port and DNS choices.
-	// For now, assume they are the same for unit tests.
-	return nodeName
+	// If the nodeName already contains a port (e.g. "127.0.0.1:12345" in unit tests), use it as-is.
+	if strings.Contains(nodeName, ":") {
```
aishukamal: This would break for IPv6 addresses. Consider using `net.SplitHostPort` which handles both IPv4 and IPv6.

pkg/accelerator-orchestrator/store/snapshot_agent_store_test.go
```diff
@@ -61,7 +61,7 @@ func TestGRPCSnapshotAgentStore_GetStatus(t *testing.T) {
 	defer grpcServer.GracefulStop()
 
 	// Test with 5s TTL
-	s := store.NewGRPCSnapshotAgentStore(5 * time.Second)
+	s := store.NewGRPCSnapshotAgentStore(5*time.Second, 9001)
```
aishukamal, line 64: Nit: consider using a named constant for the default port so we don't have to update every test if this ever changes.

pkg/accelerator-orchestrator/store/snapshot_agent_store_internal_test.go
```diff
@@ -0,0 +1,69 @@
...
+	tests := []struct {
+		name        string
+		nodeName    string
+		defaultPort int
+		want        string
+	}{
+		{
+			name:        "without port",
+			nodeName:    "node-1",
+			defaultPort: 9001,
```
aishukamal, line 31: Nit: consider using a named constant for the default port so we don't have to update every test case if this ever changes.

jessicaochen: Interestingly enough, these tests are self contained (ie. it detects that the value passed in as default port is handled as expected in the resulting).

Even if we change the default in main.go, this logic will still pass. Therefore only changes in the logic of determining addresses will change these test cases.

The mass changes of concern will not happen.

pkg/accelerator-orchestrator/store/snapshot_agent_store_test.go
jessicaochen: The value passed here is arbitrary and not at all involved in what this test is verifying. It just needs to pass some sort of value. Therefore, there really should be no case where we need to change all of them for the tests to validate what they are supposed to.

pkg/accelerator-orchestrator/store/snapshot_agent_store.go
jessicaochen: Good callout, going with net package utils

pkg/accelerator-orchestrator/store/snapshot_agent_store_internal_test.go
jessicaochen: made the ports different numbers for each test case to be more obvious that the test is validating the behavior not any actual default in main.go

Issue #51, opened 2026-06-18
Make snapshot agent k8s agnostic


aishukamal:
Snapshot agent today has a k8s dependency for pod discovery. However, that dependency exists mainly because it relies on a "job-id" for workload identification. This dependency is probably fine for the internal orchestrator-snapshot agent contract but for an external service integrating directly with snapshot agent, none of these make sense. Three things worth looking into:
* Can we get rid of job_id in the API interface and use a more appropriate workload identifier? Maybe we use PID directly for backends that work on the process level.
* Once we get rid of job_id, do we still need the pod discovery logic? What does it take to remove that dependency?
* Add a user guide for non-k8s environments. See this related discussion - https://github.com/llm-d-incubation/llm-d-rl-time-slicing/pull/39#discussion_r3433379120_

aishukamal:
@lynnl0927 Since you plan on looking into the overall user onboarding journey, this could be a good starting point. This is definitely going to be a friction point for Snapshot Agent integration. Can you take a stab at it?

jessicaochen:
One clarification we should probably make is if we are not assuming k8s, are we assuming containers?

In my mind job_id is a string identifier, it can be filled with whatever is appropirate for your orchestration backend. So in it is a pod label for k8s but could easily be reused as pid for bare metal OR some sort of container id if the stuff is running in containers.

jessicaochen:
In case it helps, I made the orchestrator agnostic but putting the k8s stuff behind an interface pkg/accelerator-orchestrator/infrastructure with certain hooks.

aishukamal:
Snapshot agent now has a "standalone" mode which can run on bare-metal.

PR #58, opened 2026-06-23
Add a snapshot agent dev mode without k8s dependencies

lynnl0927:
## What does this PR do?

Add a snapshot agent dev mode without k8s dependencies

## Why is this change needed?

https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/51

## How was this tested?

## Checklist

## Related Issues

https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/51


pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
```diff
@@ -44,6 +45,7 @@ message RestoreRequest {
   // group is an optional identifier for a set of related jobs.
   string group = 2;
   Backend backend = 3;
+  repeated int32 pids = 4;
```
aishukamal: Should this be int64? I think PIDs on Linux can exceed the int32 range (though probably rare).

cmd/snapshot-agent/main.go
```diff
@@ -32,17 +32,20 @@ func main() {
 	slog.SetDefault(slog.New(ctxHandler))
 
 	port := flag.Int("port", 9001, "The port to listen on")
+	devMode := flag.Bool("dev", false, "Run in development/local mode without Kubernetes")
 	flag.Parse()
 
+	isDev := *devMode || os.Getenv("DEV_MODE") == "true"
```
aishukamal: I think we should refrain from calling this "dev" mode since non-k8s for RL is a totally valid (and perhaps the more popular) production mode. We should just call it non-k8s or something along those lines.

In fact, I'm wondering if we even need to make it a special case by having a flag at all. It could just be as simple as, if the user provides PIDs, use them, otherwise assume k8s?

Edwinhr716: If we do that, we should return an error when the user is using k8s, but still passing PIDs

aishukamal: After thinking about it for a bit, here's my mental model:

Each backend has a different checkpoint target - CUDA uses PIDs, GCR uses PIDS (?), LoRA offload uses memory addresses, etc. Rather than adding `repeated int32 pids` directly to `SnapshotRequest`, consider a more generic target abstraction (e.g., `oneof target` or backend-specific params) so each new backend doesn't require modifying the top-level proto.

With that framing, the snapshot agent's core job is: given a target, checkpoint/restore its state. How those targets are discovered is a separate concern. In standalone usage, the user provides targets directly in the request. In orchestrated usage (where the orchestrator sends a job_id without targets), the agent needs a configured discovery mechanism to resolve job_id -> targets - k8s pod lookup being the current one.

Based on this, I suggest:

* Drop the `devMode` flag. If targets are in the request, use them directly. If not, use the configured discovery mechanism.
* If a flag is needed, flip it - make discovery the opt-in (e.g., `--target-discovery=k8s`), not the default. The orchestrator's deployment would set this flag. The orchestrator itself doesn't need to change - it keeps sending job_id as today.
* For the user guide, I'd lead with direct target specification as the primary usage, then introduce k8s discovery as an optional automation layer for users who don't want to manage targets themselves and happen to run on k8s.

This makes the snapshot agent a standalone component that happens to have k8s integration, rather than a k8s component that happens to have a dev mode. Thoughts?

lynnl0927: Yeah that makes sense if we want to use the standalone agent for more than just development.

I'll make the following changes then:
1. Remove the devMode flag and replace it with a --deployment-mode flag that can be set to 'standalone' or 'k8s' and defaults to 'standalone'
2. Add an ExplicitTarget message, if specified, we'll use that field for snapshotting if possible, if not, we'll try to use the default discovery according to the deployment mode.

I'd still prefer keeping the job_id in both modes as it acts as the unique identifier for operation status tracking and restoring the checkpoint. If it's removed from the standalone mode, I think we lose some benefits of using the snapshot agent server to manage snapshot/restore actions and it would probably be easier to directly use the corresponding commands like cuda-checkpoint.

aishukamal: Sounds good to me!

pkg/client/python/timeslice/snapshot_agent/client.py
```diff
@@ -130,26 +135,55 @@ def _get_backend_enum(self, backend: Union[str, int]) -> int:
                 return snapshot_agent_pb2.BACKEND_UNSPECIFIED
 
     def snapshot(
-        self, job_id: str, group: str = "", backend: Union[str, int] = 0
+        self,
+        job_id: str,
+        group: str = "",
+        backend: Union[str, int] = 0,
+        pids: Optional[list[int]] = None,
```
Edwinhr716: I don't think it makes sense to add pids as a field here, since not all backends will have this as an option. The GCR backend for example would have memory addresses as a parameter instead. Even though it makes it more clunky for the user, I think we should have the user build the backend_config object.

lynnl0927: This might be accidentally kept from the previously abandoned commit, will remove it.

pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
aishukamal, line 70: Should this also be deprecated?

pkg/client/python/timeslice/snapshot_agent/client.py
```diff
@@ -38,7 +38,12 @@ class SnapshotAgentInterface(ABC):
 
     @abstractmethod
     def snapshot(
-        self, job_id: str, group: str = "", backend: Union[str, int] = 0
+        self,
+        job_id: str,
+        group: str = "",
+        backend: Union[str, int] = 0,
+        pids: Optional[list[int]] = None,
```
aishukamal, line 45: Is there a reason to keep this field now that we have `backend_config`? The signature is also inconsistent with the concrete implementation of this abstract method as that method doesn't have `pids` in its signature.

pkg/client/python/timeslice/snapshot_agent/client.py
```diff
@@ -130,26 +135,43 @@ def _get_backend_enum(self, backend: Union[str, int]) -> int:
                 return snapshot_agent_pb2.BACKEND_UNSPECIFIED
 
     def snapshot(
-        self, job_id: str, group: str = "", backend: Union[str, int] = 0
+        self,
+        job_id: str,
+        group: str = "",
+        backend: Union[str, int] = 0,
+        backend_config: Optional[snapshot_agent_pb2.BackendConfig] = None,
```
aishukamal, line 142: Not a blocker for this PR. I think we should file an issue to move permanently to `backend_config` and remove `backend` altogether since it adds a lot of noise in the codebase and it's too early to take on that debt. We only have 1 consumer (open-rl), so we can help open-rl make the migration and then get rid of the deprecated fields.

pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
```diff
@@ -24,13 +24,37 @@ enum Backend {
   BACKEND_CUDA = 1;
 }
 
+// ProcessTarget specifies a set of process IDs (PIDs) to target.
+// This is optional in some deployment modes (like 'k8s' where PIDs can be discovered automatically),
+// but required in others (like 'standalone').
+// In the future, we may add other target types (e.g., MemoryTarget)
+// to support different backends.
+message ProcessTarget {
+  repeated int32 pids = 1;
```
aishukamal, line 33: We probably want to use int64 to avoid overflows (although rare).

aishukamal:
/approve

(My open comments are all minor and can be addressed in a follow-up PR if needed.)

PR #62, opened 2026-06-24
Add orchestrator python client

jessicaochen:
## What does this PR do?

Adds orchestrator python client. Notice the name change of yield to release because yield is a reserved python word. Also chose to keep the client named orchestrator client because it is clear the scope of the client and what it talks to.

We *could* have a timesliceclient or timeslicer at the root level if we want. I presume it would be a client that would talk to multiple pieces eg. the orchestrator as well as the scheduler (does not exist yet).

## How was this tested?

## Checklist


pkg/client/python/timeslice/orchestrator/client.py
```diff
@@ -0,0 +1,281 @@
...
+
+try:
+    import accelerator_orchestrator_pb2 as pb2  # noqa: E402
+    import accelerator_orchestrator_pb2_grpc as pb2_grpc  # noqa: E402
+finally:
+    if _generated_dir in sys.path:
+        sys.path.remove(_generated_dir)
+
+
+class OrchestratorClient:
```
aishukamal, line 19: I've been mulling over this for a while. I think having "TimeSlice" in the name is valuable and immediately tells the user what this orchestrator is about. The original "Accelerator Orchestrator" name came from a time when we hadn't settled on whether to call this feature time-slicing, accelerator multiplexing or something else. Now that we've finalized on "timeslice," I think this should be TimeSliceOrchestratorClient or just TimeSliceClient, WDYT?

We should consider renaming the orchestrator component itself to "TimeSlice Orchestrator" too, but that's a separate discussion - let's not add too much churn right now since the component is in active development.

pkg/client/python/timeslice/orchestrator/client.py
```diff
@@ -0,0 +1,281 @@
...
+_generated_dir = str(pathlib.Path(__file__).parent / "_generated")
+if _generated_dir not in sys.path:
+    sys.path.insert(0, _generated_dir)
+
+try:
+    import accelerator_orchestrator_pb2 as pb2  # noqa: E402
+    import accelerator_orchestrator_pb2_grpc as pb2_grpc  # noqa: E402
+finally:
+    if _generated_dir in sys.path:
+        sys.path.remove(_generated_dir)
```
aishukamal: The `_generated` directory already has an `__init__.py`, so the sys.path manipulation here is unnecessary, no? Can't this be a normal relative import like below?

```
from ._generated import accelerator_orchestrator_pb2 as pb2
from ._generated import accelerator_orchestrator_pb2_grpc as pb2_grpc
```

pkg/client/python/tests/test_orchestrator.py
```diff
@@ -0,0 +1,314 @@
...
+# or we can use the same import trick.
+import pathlib
+import sys
+
+_generated_dir = str(
+    pathlib.Path(__file__).parent.parent / "timeslice" / "orchestrator" / "_generated"
+)
+if _generated_dir not in sys.path:
+    sys.path.insert(0, _generated_dir)
+import accelerator_orchestrator_pb2 as pb2  # noqa: E402
```
aishukamal: Can we just use a normal relative import? See my other comment in client.py on this.

pkg/client/python/timeslice/orchestrator/client.py
```diff
@@ -0,0 +1,281 @@
...
+        resolved_group_id = group_id or self.group_id
+
+        if not resolved_job_id:
+            raise ValueError(
+                "job_id must be provided either in the constructor or in the method call."
+            )
+        if not resolved_group_id:
+            raise ValueError(
+                "group_id must be provided either in the constructor or in the method call."
+            )
```
aishukamal: Nit: this validation looks repetitive and is duplicated across `acquire`, `release` and `get_status`. Consider extracting it into a helper function.

pkg/client/python/timeslice/orchestrator/client.py
```diff
@@ -0,0 +1,281 @@
...
+    ):
+        """Context manager for safe acquire/release flow.
+
+        Usage:
+            with client.lock():
+                # exclusive access here
+        """
+        self.acquire(job_id=job_id, group_id=group_id, timeout_sec=timeout_sec)
+        try:
+            yield
```
aishukamal: Would it make sense to yield the `AcquireResult` here? Users of the context manager form might want to inspect `waited_ms`.

pkg/client/python/timeslice/orchestrator/client.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        request = pb2.ListGroupsRequest()
+        try:
+            response = self._stub.ListGroups(request, timeout=timeout_sec)
+            return list(response.group_ids)
+        except grpc.RpcError as e:
+            raise wrap_grpc_error(e) from e
+
+    @contextlib.contextmanager
+    def lock(
```
aishukamal: I'm wondering if we should rename `lock()` to something more domain-specific - lock is generic and doesn't convey what's being locked or why. Some options:

* `on_accelerator` - describes where the code runs, reads naturally as a decorator (`@client.on_accelerator(group_id="trainer-group")`)
* `with_accelerator` - describes the precondition
* `accelerator_phase` - describes what the function represents

I'm leaning towards on_accelerator since it immediately signals "this is a GPU phase". Thoughts?

pkg/client/python/timeslice/orchestrator/_generated/accelerator_orchestrator_pb2_grpc.py
```diff
@@ -0,0 +1,245 @@
+# Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
+"""Client and server classes corresponding to protobuf-defined services."""
+import grpc
+import warnings
+
+import accelerator_orchestrator_pb2 as accelerator__orchestrator__pb2
```
aishukamal: This bare import is probably what requires the whole `sys.path` manipulation hack in `client.py`. For Snapshot Agent we solve the same problem by using [relative imports](https://github.com/llm-d-incubation/llm-d-rl-time-slicing/blob/main/pkg/client/python/timeslice/snapshot_agent/snapshot_agent_pb2_grpc.py#L6) in the generated file. Can we do the same here?

pkg/client/python/tests/test_orchestrator.py
```diff
@@ -0,0 +1,314 @@
+import datetime
```
aishukamal, line 1: Looks like CI only runs python linting but not pytest - https://github.com/llm-d-incubation/llm-d-rl-time-slicing/blob/main/.github/workflows/ci-pr-checks.yaml. Let's add a pytest step to the job so these tests actually run on PRs.

jessicaochen: Deferring this to https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/63

pkg/client/python/timeslice/orchestrator/client.py
jessicaochen: I have noted this in the description but I chose orchestrator instead of timeslice because this client is specifically about talking to the orchestrator.

I would not be shocked if we have a timeslice client that actually wraps multiple other clients to do fancier stuff calling across the orchestrator, agent, placement optimizer etc for a wholistic timeslice experience.

Would you like me to create that TimeSliceClient wrapper or just stick to OrchestratorClient in  TimeSlice "package" ?

pkg/client/python/timeslice/orchestrator/client.py
jessicaochen: I think I like on_accelerators the best, will plumb that....

pkg/client/python/timeslice/orchestrator/_generated/accelerator_orchestrator_pb2_grpc.py
jessicaochen: The price to do this is manually editing generated files (which in general is a no-no because your edits get lost on next generation): https://github.com/llm-d-incubation/llm-d-rl-time-slicing/tree/main/pkg/client/python#development

Let me see if I can find something that does not modify generated files but also has cleaner imports.....

jessicaochen: Went with modifying the __init__ in the _generated folder to take care of the sys path stuff. Now everything using it can use normal imports & generated code does not need to be manually modified each time.

pkg/client/python/tests/test_orchestrator.py
jessicaochen: addressed, see other comment.

pkg/client/python/timeslice/orchestrator/client.py
jessicaochen: addressed, see other comment.

pkg/client/python/timeslice/orchestrator/client.py
jessicaochen: Sounds good, done.

pkg/client/python/timeslice/orchestrator/client.py
jessicaochen: Makes sense, implemented.

pkg/client/python/tests/test_orchestrator.py
jessicaochen: Looks, like I can actually add the check the same PR as the unit tests so keeping it in this PR.

pkg/client/python/timeslice/orchestrator/client.py
jessicaochen: Ok, I am getting different suggestions on naming. I plan on sticking with OrchestratorClient in TimeSlice "package" for now to be flexible (and not stop development flow) till https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/64 is settled.

pkg/client/python/timeslice/orchestrator/_generated/accelerator_orchestrator_pb2_grpc.py
aishukamal: > The price to do this is manually editing generated files (which in general is a no-no because your edits get lost on next generation)

There are ways around this though. We can always write a Makefile target that generates the proto files and fixes the imports as a post-generation step. In fact, there are [tools](https://pypi.org/project/fix-protobuf-imports/) built to do exactly this since this is a well known problem. The protobuf community's suggested approach is to use `--proto_path` mappings so the generated imports match the Python package structure, avoiding the need for any post-generation fixes altogether. So something like:

`protoc --proto_path=timeslice/snapshot_agent=pkg/snapshot-agent/api/v1alpha1 ...`

The reason I don't like the `sys.path` approach is because it mutates global state to solve a local import problem. If another module in the process happens to have the same name as the generated file, the sys.path insertion could cause it to shadow or get shadowed by unrelated modules. It's unlikely in practice, but it's a class of bug that's hard to diagnose when it does happen.

pkg/client/python/timeslice/orchestrator/client.py
aishukamal: Sounds reasonable. Let's resolve this in #64.

pkg/client/python/timeslice/orchestrator/_generated/accelerator_orchestrator_pb2_grpc.py
jessicaochen: Adjusting the generation command worked to avoid sys.path manipulation, updated the docs & regenrated.

aishukamal: LGTM! Thank you for working through this!

PR #77, opened 2026-06-29
Adding pod watcher to Snapshot-Agent

Edwinhr716:
## What does this PR do?
Adds a pod watcher to the snapshot-agent, which watches whenever a pod with the job-id label is scheduled on the node where the snapshot-agent is running. If a pod is detected, the status would be set to IDLE. It then checks for GPU PIDs in the pod. Once one is detected, the status is changed to RUNNING

## How was this tested?

## Checklist


pkg/snapshot-agent/server/export_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	sm "github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/snapshot-agent/state-machine"
+)
+
+// InternalState returns the internal StateManager.
+// Exported for testing purposes only.
+func (s *Server) InternalState() *sm.StateManager {
+	return s.state
+}
```
jessicaochen: Usually if you need to access internal state  purely for testing, use whitebox tests. They seem to be named XXX_internal_test.go & are in the same package as the thing it is testing.

You usually do not want any public access to internals.

pkg/snapshot-agent/server/watcher.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func (w *Watcher) registerPodJob(pod *corev1.Pod) {
+	jobID, hasJob := pod.Labels[podutils.JobIDLabel]
+	if !hasJob {
+		return
+	}
+
+	// Group can be optional, check if there is a group label
+	group := pod.Labels["timeslice.io/group"]
```
jessicaochen: const?

pkg/snapshot-agent/server/watcher.go
```diff
@@ -0,0 +1,173 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			w.checkIdleJobs(ctx)
+		}
+	}
+}
+
+func (w *Watcher) checkIdleJobs(ctx context.Context) {
```
jessicaochen, line 132: you do not need to handle in this PR but what do we want to do when pod completes or gets deleted from the node?

jessicaochen (changes requested):

pkg/snapshot-agent/server/watcher.go
Edwinhr716: Good call out, we want to delete the associated job from the state machine

pkg/snapshot-agent/server/watcher.go
Edwinhr716: done

pkg/snapshot-agent/server/export_test.go
Edwinhr716: Changed to server_internal_test.go

pkg/snapshot-agent/server/watcher.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, fmt.Errorf("NODE_NAME environment variable not set")
+	}
+
+	// Create an informer filtered by node name
+	factory := informers.NewSharedInformerFactoryWithOptions(clientset, 0,
+		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.FieldSelector = fmt.Sprintf("spec.nodeName=%s", nodeName)
+		}),
+	)
+	podInformer := factory.Core().V1().Pods().Informer()
```
jessicaochen, line 56: (nit) you can actually have options here to reduce to timeslicing labeled pods. That way you only trigger for pods you care about as opposed to system pods on the node.

jessicaochen: Will note you still need you check later on because it could have been an update to remove the timeslice labels.

pkg/snapshot-agent/server/watcher.go
jessicaochen: I assume you plan on handling in a different PR.

jessicaochen (approved):

pkg/snapshot-agent/server/watcher.go
Edwinhr716: We don't trigger for pods, it is a loop that checks every second. The only difference it would make is that when fetching the pods, it returns a shorter list. Think it is fine to keep it this way

PR #78, opened 2026-06-29
Acquire - Refresh state & check current locker

jessicaochen:
## What does this PR do?

Fix some bugs found in integration testing with real vllm. Specifically, longer vllm snapshot meant that Acquire needed to loop longer before it got the lock. It also showed a condition where locked but not loaded state let a different lock slip through.

Did need to create whitebox internal test suite (and move all the shared mocks/fakes over) so I could use a func var to get at internal logic timing with channels instead of time.sleeps (a future source of flakes).

## How was this tested?

## Checklist

### DIFF
--- a/pkg/accelerator-orchestrator/server/server.go
+++ b/pkg/accelerator-orchestrator/server/server.go
@@ -30,23 +30,29 @@ type JobStore interface {
 	ListByGroup(ctx context.Context, groupID string) ([]*store.Job, error)
 }
 
-const acquirePollInterval = 1 * time.Second
+// checkAcquireFunc defines the signature for the Acquire check hook.
+type checkAcquireFunc func(ctx context.Context, groupID, jobID string, startTime time.Time) (*pb.AcquireResponse, error, bool)
 
 // Server implements the AcceleratorOrchestratorService gRPC server.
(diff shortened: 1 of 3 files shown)


pkg/accelerator-orchestrator/server/server.go
```diff
@@ -85,28 +91,48 @@ func (s *Server) Acquire(ctx context.Context, req *pb.AcquireRequest) (*pb.Acqui
 			slog.InfoContext(ctx, "Acquire context cancelled", "error", ctx.Err())
 			return nil, status.FromContextError(ctx.Err()).Err()
```
lynnl0927, line 92: If we just return here, could there be an issue where the lock is acquired but the request is cancelled, and then the lock is never released? Not blocking this.

lynnl0927 (approved):

pkg/accelerator-orchestrator/server/server.go
jessicaochen: You are right, I think I need to either release the lock in this case. Will break out to separate task.

https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/80

PR #84, opened 2026-07-08
Clean up backend enum in snapshot agent client

lynnl0927:
## What does this PR do?

This PR finalizes the migration from the legacy `Backend` enum to the new `BackendConfig` structure in the Snapshot Agent API.

The server now completely ignores the legacy `backend` enum field. All internal routing and logic have been updated to strictly depend on `BackendConfig`.

Action Required for Users: If you are calling the Snapshot Agent directly (e.g., via grpcurl or older client versions), you must update your requests to specify `backend_config` instead of `backend`.

## Why is this change needed?

https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/51

## How was this tested?

## Checklist

- [x] Tests pass locally (`make test`)
- [x] Linters pass (`make lint`)
- [x] Documentation updated (if applicable)

## Related Issues

https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/51

## Summary by CodeRabbit

* **New Features**
  * Snapshot and restore now support backend-specific configuration via `backend_config` (including CUDA), in addition to legacy backend selection.
* **Bug Fixes**
  * Restore backend selection now follows the same `backend_config`-first behavior as snapshot.
  * Legacy backend selection remains supported for backward compatibility (now deprecated).
* **Documentation**
  * Updated the Python client README and usage examples to use `backend_config`.
* **Tests**
  * Expanded server-side tests to validate `backend_config`-based snapshot and restore requests.

### DIFF
--- a/pkg/accelerator-orchestrator/store/snapshot_agent_store.go
+++ b/pkg/accelerator-orchestrator/store/snapshot_agent_store.go
@@ -164,9 +164,8 @@ func (s *GRPCSnapshotAgentStore) Restore(
 	}
 
 	resp, err := client.Restore(ctx, &agentpb.RestoreRequest{
-		JobId:   jobID,
-		Group:   groupID,
-		Backend: agentpb.Backend_BACKEND_CUDA, // Default to CUDA
+		JobId: jobID,
+		Group: groupID,
 	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to trigger restore for job %s on agent at %s: %w", jobID, address, err)
--- a/pkg/client/python/timeslice/snapshot_agent/client.py
+++ b/pkg/client/python/timeslice/snapshot_agent/client.py
@@ -1,7 +1,7 @@
 import logging
 import time
 from abc import ABC, abstractmethod
-from typing import Optional, Union
+from typing import Optional
 
 import grpc
 
@@ -41,8 +41,6 @@ class SnapshotAgentInterface(ABC):
         self,
         job_id: str,
         group: str = "",
-        backend: Union[str, int] = 0,
-        pids: Optional[list[int]] = None,
         backend_config: Optional[snapshot_agent_pb2.BackendConfig] = None,
     ) -> SnapshotResponse:
         """Triggers an asynchronous snapshot."""
@@ -50,7 +48,10 @@ class SnapshotAgentInterface(ABC):
 
     @abstractmethod
     def restore(
-        self, job_id: str, group: str = "", backend: Union[str, int] = 0
+        self,
+        job_id: str,
+        group: str = "",
+        backend_config: Optional[snapshot_agent_pb2.BackendConfig] = None,
     ) -> RestoreResponse:
         """Triggers an asynchronous restoration."""
         pass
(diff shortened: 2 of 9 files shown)


pkg/accelerator-orchestrator/store/snapshot_agent_store.go
jessicaochen, line 125: (sanity check) how come snapshot does not need defaulting here but restore does? I would have expected consistency (eg. they both need defaulting or they both do not need defaulting here).

jessicaochen (dismissed):

pkg/snapshot-agent/server/server.go
```diff
@@ -126,9 +127,14 @@ func (s *Server) Restore(ctx context.Context, req *pb.RestoreRequest) (*pb.Resto
...
-	slog.InfoContext(ctx, "Restore called", "backend", req.GetBackend())
-
-	backendType := s.getBackendType(req.GetBackend())
+	var backendType backends.BackendType
+	if req.GetBackendConfig() != nil {
+		backendType = s.getSnapshotBackendType(req.GetBackendConfig())
+	} else {
+		//nolint:staticcheck // SA1019: Keep supporting deprecated backend field for backward compatibility
+		backendType = s.getBackendType(req.GetBackend())
+	}
```
aishukamal: Is this still needed? We might have external users using the deprecated `Backend` field but all the entrypoints (python client and orchestrator go client) have already migrated to `BackendConfig` which allows the backend code to operate purely on `BackendConfig`, no?

pkg/client/python/timeslice/snapshot_agent/client.py
```diff
@@ -117,61 +118,31 @@ def _handle_rpc_error(self, e: grpc.RpcError, method_name: str):
...
-                logger.warning(
-                    f"Unknown backend '{backend}', using BACKEND_UNSPECIFIED"
-                )
-                return snapshot_agent_pb2.BACKEND_UNSPECIFIED
-
     def snapshot(
         self,
         job_id: str,
         group: str = "",
-        backend: Union[str, int] = 0,
```
aishukamal, line 141: Wouldn't this break open-rl? We need to make sure open-rl has migrated off of this field before removing it from the client. Until then, we continue to support the deprecated field on the client RPCs but internally we migrate the codebase away to `backend_config` completely. So the client here can simply construct `backend_config` from `backend` and pass it down further. All downstream code should expect only `backend_config`. WDYT?

aishukamal (changes requested):

pkg/snapshot-agent/server/server.go
aishukamal: Same here. We can get rid of the `else` block if we ensure all entrypoints construct a `BackendConfig` and plumb it down.

We should also check the codebase for any other users of the `Backend` field and remove it.

aishukamal (changes requested):

pkg/accelerator-orchestrator/store/snapshot_agent_store.go
lynnl0927: Synced offline, removing the default consistency.

pkg/client/python/timeslice/snapshot_agent/client.py
lynnl0927: Confirmed offline that open-rl changes are still on a feature branch, and they have control on what version of snapshot agent to ship with, we don't need to worry about backward compatibility for now.

pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
```diff
@@ -67,7 +68,10 @@ message RestoreRequest {
   string job_id = 1;
   // group is an optional identifier for a set of related jobs.
   string group = 2;
-  Backend backend = 3;
+  // Deprecated: Use backend_config instead.
+  Backend backend = 3 [deprecated = true];
```
aishukamal, line 72: Nit (not blocking): we can technically remove this as well, instead of marking deprecated, no? Since we don't have any users currently, it would be good to start from a clean base.

aishukamal (approved):

PR #85, opened 2026-07-09
Crash Recovery For Mid Transitioning

jessicaochen:
Logic to resume work when the orchestrator gets restarted while active jobs is in a transitioning action.

## Why is this change needed?

We want the orchestrator to be able to resume work if it is ever rebooted during normal operations.

## How was this tested?

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved controller behavior when the active job is marked as transitioning, preventing premature reconciliation and retrying the work item later.
  * Added an active-job-specific safety gate so restore/preemption actions are skipped until the transition completes.

* **Tests**
  * Added a new test covering the transitioning active-job scenario, verifying reconciliation fails and the item is re-queued without triggering unexpected snapshot/restore calls.

### DIFF
--- a/pkg/accelerator-orchestrator/controller/controller.go
+++ b/pkg/accelerator-orchestrator/controller/controller.go
@@ -283,11 +283,9 @@ func (c *Controller) reconcileNode(ctx context.Context, groupID, nodeName, activ
 		if state, ok := agentJobStates[activeJobID]; ok && state == pb.SnapshotAgentJobState_STATE_FAULTED {
 			return fmt.Errorf("active job %s is in FAULTED state on node %s, requires human intervention", activeJobID, nodeName)
 		}
-	}
 
-	// Wait out any existing transitioning. We lack the operation ID to track so wait via error
-	// to retry later via rate-limited requeue.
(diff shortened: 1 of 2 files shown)


pkg/accelerator-orchestrator/controller/controller.go
```diff
@@ -283,9 +283,14 @@ func (c *Controller) reconcileNode(ctx context.Context, groupID, nodeName, activ
 		}
 	}
 
-	// 2. Wait out any existing transitions
-	// TODO: crash recovery detect that an agent is still in the middle of transitioning. We will not have the operation
-	// number but we can still wait till it leaves that state.
+	// Wait out any existing transitioning. We lack the operation ID to track so wait via error
+	// to retry later via rate-limited requeue.
+	if activeJobID != "" {
```
aishukamal: Can this be folded into the early exits `if` block above? It is another early exit condition, after all.

jessicaochen: That makes sense, done :)

PR #86, opened 2026-07-09
Crash recovery for STATE_IDLE_YIELDED

jessicaochen:
## What does this PR do?

Adds the best effort ability to recover after a reboot when the state is having a job in STATE_IDLE_YIELDED. In other words, no one is holding the lock & the last yielder remains active till a new and different job asks for the lock.

## How was this tested?

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restart handling so the controller can recover the active job more reliably when state is missing or incomplete.
  * Reconciliation now uses a consistent active job selection when updating node and group status, reducing chances of inconsistent state after a restart.

* **Tests**
  * Added coverage for restart scenarios to verify active job detection and the resulting group status across multiple cases.

### DIFF
--- a/pkg/accelerator-orchestrator/controller/controller.go
+++ b/pkg/accelerator-orchestrator/controller/controller.go
@@ -220,30 +220,32 @@ func (c *Controller) reconcileGroup(ctx context.Context, groupID string) error {
 		return fmt.Errorf("failed to observe job context: %w", err)
 	}
 
-	// TODO: deduce activejob for restart case when group is in STATE_IDLE_YIELDED
-
 	// 2. Determine Desired State
-	g, err := c.groupStore.Get(ctx, groupID)
+	group, err := c.groupStore.Get(ctx, groupID)
 	if err != nil {
 		return fmt.Errorf("failed to get group %s from store: %w", groupID, err)
 	}
 
-	if _, err := g.Spec().TryPromote(ctx); err != nil {
+	if err := c.tryDeduceActiveJob(ctx, group); err != nil {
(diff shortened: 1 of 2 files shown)


aishukamal:
@coderabbitai review

pkg/accelerator-orchestrator/controller/controller.go
```diff
@@ -343,6 +345,44 @@ func (c *Controller) reconcileNode(ctx context.Context, groupID, nodeName, activ
...
+	}
+
+	jobs, err := c.jobStore.ListByGroup(ctx, group.ID())
+	if err != nil {
+		return fmt.Errorf("failed to list jobs for group %s: %w", group.ID(), err)
+	}
+
+	var loadedJob string
+	for _, job := range jobs {
+		loaded, err := c.isJobLoaded(ctx, group, job.JobID())
```
aishukamal, line 363: Nit and non-blocking: `tryDeduceActiveJob` calls `isJobLoaded` in a loop and each call does its own `ListByGroup` and rebuilds the node -> running-job map from scratch, so for N jobs we do N+1 redundant list+map builds. Fine at current scale, but consider splitting `isJobLoaded` into a `buildNodeRunningMap` helper (called once) plus a cheap per-job check against the pre-built map. `tryDeduceActiveJob` could then also reuse the jobs list from the map build instead of its own `ListByGroup`.

jessicaochen: Agreed this will be important later when scalability is more important. Made tracker Issue

PR #87, opened 2026-07-09
Orc metrics

jessicaochen:
## What does this PR do?

MVP set of metrics for the orchestrator.

## How was this tested?

- [x] Manual testing performed - I do see all the metrics fill out

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for queue depth, lock acquire wait time, snapshot/restore operation duration, and deferred snapshots.
  * Exposed a dedicated HTTP `/metrics` endpoint on a configurable `--metrics-port` (default 8080), including Kubernetes service/deployment updates.
* **Bug Fixes**
  * Improved graceful shutdown by stopping the metrics HTTP server with a timeout.
  * Refined metric recording to align with acquire/yield and reconciliation timing.
  * Cleaned up stale queue-depth metric series when groups are deleted.
* **Tests**
  * Expanded metric assertions for queue depth transitions, acquire wait observations, operation-duration histograms, and deferred-snapshot counters.


pkg/accelerator-orchestrator/metrics/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+		},
+		[]string{"group_id"},
+	)
+
+	// AgentOperationDuration tracks the duration of snapshot and restore operations as reported by snapshot agents.
+	AgentOperationDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "accelerator_orchestrator_agent_operation_duration_seconds",
+			Help:    "Duration of snapshot and restore operations as reported by snapshot agents.",
+			Buckets: prometheus.DefBuckets,
```
aishukamal: The bucket granularity in `DefBuckets` range from [5ms to 10s](https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L271) which won't for `AcquireWaitDuration` or `AgentOperationDuration`. We probably need to define customized buckets.

pkg/accelerator-orchestrator/metrics/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+		prometheus.CounterOpts{
+			Name: "accelerator_orchestrator_deferred_snapshots_total",
+			Help: "Number of times a snapshot was deferred during Yield() due to zero waiters.",
+		},
+		[]string{"group_id"},
+	)
+)
+
+// Register registers all accelerator orchestrator Prometheus metrics with the default registry.
+func Register() {
```
aishukamal, line 60: Do we need a cleanup function too? When a group is deleted I believe the last emitted value persists in `/metrics` forever and could be misleading, especially for `QueueDepth`.

pkg/accelerator-orchestrator/server/server.go
```diff
@@ -274,6 +282,21 @@ func StartServer(
...
 	}
 
+	// Start HTTP metrics server
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+	httpServer := &http.Server{
+		Addr:              fmt.Sprintf(":%d", metricsPort),
+		Handler:           mux,
+		ReadHeaderTimeout: 5 * time.Second,
+	}
```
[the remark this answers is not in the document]
aishukamal: The metrics port is cluster-internal (not exposed via ingress), serving only Prometheus scrapes, so the threat model for a Slowloris attack on it is weak. But the fix is three lines and standard practice for any http.Server, so we should probably implement it anyway.

@jessicaochen

aishukamal (changes requested):

pkg/accelerator-orchestrator/server/server.go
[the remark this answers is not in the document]
jessicaochen: Added the timeouts.

pkg/accelerator-orchestrator/metrics/metrics.go
jessicaochen: Good point, updated.

pkg/accelerator-orchestrator/metrics/metrics.go
jessicaochen: Yes, I specifically added cleanup for the queue depth on delete.

I did leave the other ones alone mostly thinking for a future node autoscalar case. Can revisit later if we do not persue that feature & holding onto group historical metrics gets too heavy.

aishukamal (approved):

PR #88, opened 2026-07-10
snapshot-agent: bootstrap job state on first snapshot in standalone mode

aishukamal:
## Summary

In standalone mode there is no watcher to move jobs out of IDLE, so a first `Snapshot` request was always rejected. On `Snapshot`, the server now registers the job and — if it is IDLE while the GPU has running compute processes (NVML check) — transitions it to RUNNING.

This applies in **standalone mode only**. Discovery does two things: kicking off the job state machine, and binding jobs to their targets (e.g. PIDs for the CUDA backend). In k8s mode the watcher owns the first for all backends; the second is backend-specific and will move behind a per-backend discovery interface.

## Testing

Unit tests cover the standalone auto-transition (GPU occupied and unoccupied) and that k8s mode does not bootstrap watcher-unknown jobs. `go build`, `go test`, golangci-lint: clean.

## Summary by CodeRabbit

* **New Features**
  * Snapshot requests can now detect active GPU compute workloads and automatically transition the associated job to a running state.
  * CUDA snapshots in standalone mode succeed when GPU compute activity is detected (including when explicit PID targeting is provided).
  * Behavior is gated by server mode, so unsupported requests fail appropriately when GPU is occupied.

* **Bug Fixes**
  * CUDA snapshot requests now fail with a clear precondition error when no GPU workload is active.

* **Tests**
  * Added an in-memory gRPC test harness and coverage for GPU-occupied vs non-occupied snapshot behavior.

### DIFF
--- a/pkg/snapshot-agent/server/server.go
+++ b/pkg/snapshot-agent/server/server.go
@@ -70,6 +70,8 @@ func (s *Server) Snapshot(ctx context.Context, req *pb.SnapshotRequest) (*pb.Sna
 		return nil, status.Errorf(codes.NotFound, "backend %s not found", backendType)
 	}
 
+	s.ensureJobRunningIfGPUOccupied(ctx, req.GetJobId(), req.GetGroup())
+
 	bgCtx := context.WithoutCancel(ctx)
 	opID, err := s.state.StartSnapshot(req.GetJobId(), req.GetGroup(), func() error {
 		slog.InfoContext(bgCtx, "Background: Starting snapshot", "backend", backendType)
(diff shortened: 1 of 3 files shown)


pkg/snapshot-agent/server/server.go
```diff
@@ -105,6 +107,37 @@ func (s *Server) getSnapshotBackendType(config *pb.BackendConfig) backends.Backe
...
 }
 
+// ensureJobRunningIfGPUOccupied registers the job and, if it is IDLE while
+// the GPU has running compute processes, transitions it to RUNNING.
+//
+// Standalone mode only: in k8s mode the watcher is the single source of
+// state-machine transitions (and additionally binds jobs to their targets,
+// e.g. PIDs for the CUDA backend — a backend-specific concern that a future
+// discovery interface will own per backend).
+func (s *Server) ensureJobRunningIfGPUOccupied(ctx context.Context, jobID, group string) {
```
lynnl0927, line 117: do we need a per backend implementation in the future?

lynnl0927 (approved):

pkg/snapshot-agent/server/server.go
aishukamal: Yes, we will. That is the "discovery interface" mentioned in the comment above. I plan to get that done in a separate PR.

PR #92, opened 2026-07-10
Slime integration guide with runnable example

jessicaochen:
## What does this PR do?

Adds a guide with a working example integration customers can run

## How was this tested?

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive Time-Slicing Integration Guide for Slime workloads, including Kubernetes/GPU prerequisites, Helm deployment, monitoring, and troubleshooting.
  * Added a multi-cluster, time-sliced disaggregated GRPO example with Kubernetes DRA resource-claim setup, concurrent job launch, monitoring, and cleanup.
  * Included ready-to-use templates and scripts for RayJob configuration, GPU resource claims, and node/container initialization.
  * Updated accelerator orchestrator Kubernetes prerequisite to version 1.34+ for DRA v1.

### DIFF
--- a/guides/rl-frameworks/slime/README.md
+++ b/guides/rl-frameworks/slime/README.md
@@ -27,6 +27,9 @@ Before deploying cooperative time-slicing for Slime, ensure your Kubernetes clus
 ### Kubernetes Version
 * Kubernetes **v1.34** or later.
 
+### KubeRay Version
+* **KubeRay Operator v1.6.0 or later** (required for modern Dynamic Resource Allocation pod syntax).
+
 ### GPU Node Configuration
 * GPU nodes must run **NVIDIA GPU Driver 565 or later**. This is a strict requirement to support **NVIDIA Dynamic Resource Allocation (DRA)**.
 * GPU memory capacity must be sufficient to hold the active working set of a single Slime job's trainer or sampler at any one time (since inactive jobs will have their GPU memory checkpointed and evicted).
--- a/guides/rl-frameworks/slime/README.md
+++ b/guides/rl-frameworks/slime/README.md
@@ -1,0 +1,385 @@
+# Time-Slicing Integration Guide for Slime Workloads
+
+This guide provides step-by-step instructions on how to integrate and deploy **Slime** (high-performance RL framework for LLMs) with the **llm-d-rl-time-slicing** platform.
+
+### Motivation: Maximizing GPU Utilization
+In traditional disaggregated RL setups, GPUs sit idle whenever worker groups wait for another phase to complete (e.g., trainer GPUs idling during rollout generation, or rollout GPUs idling during policy updates). Cooperative time-slicing enables multiple independent Slime jobs to multiplex physical GPU resource pools concurrently. When one job finishes a phase, its GPU context is checkpointed and evicted, allowing another job to immediately utilize the hardware—significantly driving up GPU duty cycle and overall cluster throughput.
+
+For a step-by-step reproduction guide using RayCluster Kubernetes manifests and disaggregated GRPO launch scripts, see:
+* **[Multi-Cluster Time-Sliced Disaggregated GRPO Guide](sync/README.md)**
+
+---
+
+## Table of Contents
+1. [Cluster Prerequisites](#1-cluster-prerequisites)
+2. [Deploying the Time-Slicing Platform](#2-deploying-the-time-slicing-platform)
+3. [Code Integration with Slime](#3-code-integration-with-slime)
+4. [Deploying the Modified Slime Variant](#4-deploying-the-modified-slime-variant)
+5. [Submitting and Observing Time-Sliced Jobs](#5-submitting-and-observing-time-sliced-jobs)
+6. [Observing Convergence and Job Completion](#6-observing-convergence-and-job-completion)
+
+---
+
+## 1. Cluster Prerequisites
+
+Before deploying cooperative time-slicing for Slime, ensure your Kubernetes cluster meets the following requirements:
+
+### Kubernetes Version
+* Kubernetes **v1.26** or later.
+
+### GPU Node Configuration
+* GPU nodes must run **NVIDIA GPU Driver 565 or later**. This is a strict requirement to support **NVIDIA Dynamic Resource Allocation (DRA)**, which enables transparent context switching and snapshot/restore of GPU state.
+* GPU memory capacity must be sufficient to hold the active working set of a single Slime job's trainer or sampler at any one time (since inactive jobs will have their GPU states checkpointed and evicted).
+
+### Node Labeling for Time-Slice Pools
+The `timeslice` platform relies on node labels to identify resource pools (groups). For disaggregated Slime executions, label your GPU nodes accordingly:
+* **Sampler Nodes**:
+  ```bash
+  kubectl label nodes <node-name> group.timeslice.io/samplers=true
+  ```
+* **Trainer Nodes**:
+  ```bash
+  kubectl label nodes <node-name> group.timeslice.io/trainers=true
+  ```
+
+---
+
+## 2. Deploying the Time-Slicing Platform
+
+We deploy the core platform components—**Accelerator Orchestrator**, **Snapshot Agent** (DaemonSet), and the **NVIDIA DRA Driver**—using the parent Helm chart.
+
+### Step 1: Update Helm Chart Dependencies
+From the root of your `llm-d-rl-time-slicing` workspace, navigate to the `deploy` directory and fetch the required subcharts:
+```bash
+cd deploy/
+helm dependency update .
+```
+
+### Step 2: Configure `values.yaml`
+Review or modify the parent `values.yaml` file to match your cluster environment:
+```yaml
+acceleratororchestrator:
+  replicaCount: 2
+  image:
+    tag: latest
+
+snapshot-agent:
+  image:
+    tag: latest
+
+nvidia-dra-driver-gpu:
+  enabled: true
+  # Use "/home/kubernetes/bin/nvidia/" for GKE Container-Optimized OS (COS) nodes.
+  # Use "/opt/nvidia" for standard Ubuntu/Debian nodes.
+  nvidiaDriverRoot: "/home/kubernetes/bin/nvidia/"
+```
+
+### Step 3: Install the Helm Chart
+Install the chart into a dedicated namespace (`timeslice-system`). This ensures all service accounts, RBAC policies, and daemons are securely isolated:
+```bash
+helm install timeslice . -n timeslice-system --create-namespace
+```
+
+### Step 4: Verify Platform Health
+Verify that the orchestrator and agents are running and healthy:
+1. **Using the `rlts` CLI**:
+   Build the CLI tool and run the verify command:
+   ```bash
+   go build -o bin/rlts ./cmd/rlts
+   ./bin/rlts orchestrator verify
+   ```
+2. **Using kubectl**:
+   Ensure all pods in the `timeslice-system` namespace are `Running`:
+   ```bash
+   kubectl get pods -n timeslice-system
+   ```
+
+---
+
+## 3. Code Integration with Slime
+
+To participate in cooperative time-slicing, the Slime training loop driver requests and yields access to the GPU resource pools at its natural phase boundaries.
+
+Because worker processes (SGLang engines and Megatron-LM trainer actors) run as background servers, **only the main RL loop driver script (`train.py`)** needs to communicate with the Accelerator Orchestrator via `OrchestratorClient`.
+
+### Step 1: Add Time-Slicing Command-Line Arguments
+Add time-slicing configuration options to `slime/utils/arguments.py`:
+
+```python
+parser.add_argument(
+    "--enable-timeslice",
+    action="store_true",
+    default=False,
+    help="Enable llm-d-rl-time-slicing cooperative GPU grant acquisition.",
+)
+parser.add_argument(
+    "--timeslice-orchestrator-addr",
+    type=str,
+    default="timeslice-acceleratororchestrator.timeslice-system.svc.cluster.local:50051",
+    help="Address of the Accelerator Orchestrator gRPC service.",
+)
+parser.add_argument(
+    "--timeslice-job-id",
+    type=str,
+    default=None,
+    help="Unique job identifier for the Accelerator Orchestrator.",
+)
+parser.add_argument(
+    "--timeslice-sampler-group",
+    type=str,
+    default="group-slime-sampler",
+    help="Accelerator Orchestrator time-slice group for rollout samplers.",
+)
+parser.add_argument(
+    "--timeslice-trainer-group",
+    type=str,
+    default="group-slime-trainer",
+    help="Accelerator Orchestrator time-slice group for trainer actors.",
+)
+```
+
+### Step 2: Initialize Client & Allocate Placement Groups (`train.py`)
+In `train.py`, instantiate clients for both the sampler and trainer GPU groups. To prevent cross-cluster circular wait deadlocks, enforce a **Trainer-First lock hierarchy**: acquire the Trainer lock before requesting placement groups from Ray, and create actor and rollout groups concurrently using a `ThreadPoolExecutor`:
+
+```python
+import os, concurrent.futures
+from timeslice import OrchestratorClient
+
+def train(args):
+    sampler_client = None
+    trainer_client = None
+    job_id = getattr(args, "timeslice_job_id", None) or os.getenv("TIMESLICE_JOB_ID", "slime-job-default")
+
+    if getattr(args, "enable_timeslice", False):
+        addr = getattr(args, "timeslice_orchestrator_addr", "timeslice-acceleratororchestrator.timeslice-system.svc.cluster.local:50051")
+        sampler_client = OrchestratorClient(target=addr, job_id=job_id, group_id=getattr(args, "timeslice_sampler_group", "group-slime-sampler"))
+        trainer_client = OrchestratorClient(target=addr, job_id=job_id, group_id=getattr(args, "timeslice_trainer_group", "group-slime-trainer"))
+
+    # DEADLOCK PREVENTION: Acquire Trainer grant first, then Sampler.
+    if trainer_client:
+        trainer_client.acquire()
+
+    def _create_rollout_group():
+        if sampler_client:
+            sampler_client.acquire()
+        return create_placement_groups(args, role="rollout")
+
+    # Parallel deployment of the actor (ie. trainer) & sampler groups    
+    with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+        f_actor = executor.submit(create_placement_groups, args, role="actor") 
+        f_rollout = executor.submit(_create_rollout_group)
+        pgs = f_actor.result()
+        pgs.update(f_rollout.result())
+```
+
+### Step 3: Wrap Rollout and Training Phases
+Acquire and release GPU grants around the rollout collection and policy training loops in `train.py`. Note that because weight synchronization (`update_weights`) executes via direct GPU-to-GPU NCCL broadcast, **both Trainer and Sampler locks must be held concurrently** during the transfer:
+
+```python
+    # Yield Trainer lock after initial weight sync, retaining Sampler lock for rollout
+    if trainer_client:
+        trainer_client.release()
+
+    for rollout_id in range(args.start_rollout_id, args.num_rollout):
+        # ---------------------------------------------------------
+        # Phase 1: Rollout Generation (Rollout GPU Group)
+        # ---------------------------------------------------------
+        rollout_data_ref = ray.get(rollout_manager.generate.remote(rollout_id))
+        if sampler_client:
+            sampler_client.release()
+
+        # ---------------------------------------------------------
+        # Phase 2: Megatron-LM Policy Training (Trainer GPU Group)
+        # ---------------------------------------------------------
+        if trainer_client:
+            trainer_client.acquire()
+
+        actor_model.async_train(rollout_id, rollout_data_ref)
+        
+        # Re-acquire Sampler lock (following Trainer-first order) for GPU-to-GPU NCCL broadcast
+        if sampler_client:
+            sampler_client.acquire()
+
+        actor_model.update_weights()
+
+        # Yield Trainer lock, retaining Sampler lock for next rollout epoch
+        if trainer_client:
+            trainer_client.release()
+
+    if sampler_client:
+        sampler_client.release()
+        sampler_client.close()
+    if trainer_client:
+        trainer_client.close()
+```
+
+> [!NOTE]
+> **Cooperative Memory Offloading is Optional:** Application-level offloading (`--offload-train` and `--offload-rollout`) is optional. The node-local Snapshot Agent can transparently context-switch the entire GPU VRAM state without code changes. Cooperative offloading simply optimizes PCIe swap latency by discarding transient SGLang KV caches and copying only static model weights (~1GB) to CPU memory before yielding locks.
+
+> [!TIP]
+> For a tested codebase reference branch containing these exact changes, see [jessicaochen/slime (timeslice branch)](https://github.com/jessicaochen/slime/tree/timeslice).
+> For a detailed walkthrough of all codebase changes made in this fork (categorized by scheduling, device leak fixes, and memory offloading), see **[Detailed Fork Changes](sync/SLIME_CHANGES.md)**.
+
+---
+
+## Concrete Deployment & Benchmark Reference
+
+For a step-by-step reproduction guide using RayCluster Kubernetes manifests, disaggregated GRPO launch scripts, and verified performance benchmark numbers, see:
+
+* **[Disaggregated Time-Sliced Slime Deployment Guide](sync/README.md)**
+
+---
+
+## 4. Deploying the Modified Slime Variant
+
+To run your modified Slime workload on the cluster, you must package the `timeslice` client library and configure the Kubernetes deployments.
+
+### Step 1: Package and Containerize
+Ensure the `timeslice` Python client is installed in your Slime container image. Add the following to your Slime `Dockerfile`:
+
+<!-- TDB: Less than 98% confident in the exact base image or Dockerfile structure of the Slime workload. Customize this step to fit your existing Docker build process. -->
+```dockerfile
+# Copy the local timeslice Python client library into the image
+COPY pkg/client/python /opt/timeslice-client
+
+# Install the client library and its dependencies (grpcio, protobuf, etc.)
+RUN pip install /opt/timeslice-client
+```
+
+### Step 2: Configure KubeRay `RayJob` with DRA Resource Claims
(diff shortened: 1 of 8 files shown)


guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+6. [Observing Convergence and Job Completion](#6-observing-convergence-and-job-completion)
+
+---
+
+## 1. Cluster Prerequisites
+
+Before deploying cooperative time-slicing for Slime, ensure your Kubernetes cluster meets the following requirements:
+
+### Kubernetes Version
+* Kubernetes **v1.26** or later.
```
aishukamal: The versions here says v1.26 but the ResourceClaims use `resource.k8s.io/v1` (GA DRA, 1.34+) and the platform guide says 1.32+. I suggest aligning with guides/accelerator-orchestrator/README.md instead of stating a separate version here.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+### GPU Node Configuration
+* GPU nodes must run **NVIDIA GPU Driver 565 or later**. This is a strict requirement to support **NVIDIA Dynamic Resource Allocation (DRA)**, which enables transparent context switching and snapshot/restore of GPU state.
```
aishukamal: It's a bit confusing that you mention DRA here. IIUC, the driver version is a requirement strictly for checkpoint/restore and has nothing to do with DRA.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+* GPU memory capacity must be sufficient to hold the active working set of a single Slime job's trainer or sampler at any one time (since inactive jobs will have their GPU states checkpointed and evicted).
```
aishukamal: Do we also need to mention the host memory needs to be sufficient to hold the memory footprint of N trainers/samplers?

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Node Labeling for Time-Slice Pools
```
aishukamal: What about the `timeslice.io/enabled=true` label and the `timeslice.io/shared=true:NoSchedule` taint used in the template?

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Step 2: Configure `values.yaml`
+Review or modify the parent `values.yaml` file to match your cluster environment:
+```yaml
+acceleratororchestrator:
+  replicaCount: 2
+  image:
+    tag: latest
+
+snapshot-agent:
```
aishukamal, line 114: Do we need the `DEPLOYMENT_MODE=k8s` env var set?

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Step 1: Add Time-Slicing Command-Line Arguments
+Add time-slicing configuration options to `slime/utils/arguments.py`:
+
+```python
+parser.add_argument(
+    "--enable-timeslice",
+    action="store_true",
+    default=False,
+    help="Enable llm-d-rl-time-slicing cooperative GPU grant acquisition.",
```
aishukamal: Nit: s/GPU grant/accelerator lock. Consider using "accelerator" instead of "GPU" where applicable.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+> [!TIP]
+> For a tested codebase reference branch containing these exact changes, see [jessicaochen/slime (timeslice branch)](https://github.com/jessicaochen/slime/tree/timeslice).
+> For a detailed walkthrough of all codebase changes made in this fork (categorized by scheduling, device leak fixes, and memory offloading), see **[Detailed Fork Changes](sync/SLIME_CHANGES.md)**.
+
+---
+
+## Concrete Deployment & Benchmark Reference
+
+For a step-by-step reproduction guide using RayCluster Kubernetes manifests, disaggregated GRPO launch scripts, and verified performance benchmark numbers, see:
```
aishukamal: This links to sync/README for "verified performance benchmark numbers," but there are none there. I suggest adding a small table (per-job step time solo vs. 2-job, duty cycle comparison) or removing this from text.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+<!-- TDB: Less than 98% confident in the exact base image or Dockerfile structure of the Slime workload. Customize this step to fit your existing Docker build process. -->
+```dockerfile
+# Copy the local timeslice Python client library into the image
+COPY pkg/client/python /opt/timeslice-client
+
+# Install the client library and its dependencies (grpcio, protobuf, etc.)
+RUN pip install /opt/timeslice-client
+```
+
+### Step 2: Configure KubeRay `RayJob` with DRA Resource Claims
```
aishukamal, line 289: I suggest making this section the complete "time-slicing delta". This guide should be fully self sufficient so readers don't have to reverse-engineer the manifests in sync/. See my comments below on the specific gaps.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+When deploying Slime across independent Ray clusters, use KubeRay `RayJob` manifests configured with **Kubernetes Dynamic Resource Allocation (DRA)** (`resourceClaims`). Bounding containers to shared DRA claims (`shared-trainers-gpu-claim` and `shared-samplers-gpu-claim`) instead of static `nvidia.com/gpu` limits allows multiple jobs' worker pods to co-locate on the same physical GPU nodes without scheduler blocking:
+
+```yaml
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: slime-job-a
+spec:
+  rayClusterSpec:
+    workerGroupSpecs:
```
aishukamal, line 305: The fork's placement groups request custom resources (`{"trainers": 1}`), which only exist if worker groups declare them in rayStartParams. Consider adding them to this snippet.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+    - groupName: trainer-group
+      template:
+        metadata:
+          labels:
+            timeslice.io/group: trainers
+            timeslice.io/job-id: slime-job-a
+        spec:
+          nodeSelector:
```
aishukamal, line 315: I think we're missing tolerations `nvidia.com/gpu` and `timeslice.io/shared`. As written these pods won't schedule on the tainted nodes.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+            env:
+            - name: TIMESLICE_JOB_ID
+              value: "slime-job-a"
+            resources:
+              claims:
+              - name: gpu-claim
+            lifecycle:
+              postStart:
+                exec:
+                  command: ["/bin/sh", "-c", "/opt/scripts/setup_node.sh"]
```
aishukamal: This seems specific to the example in sync/ and not functional here. I suggest dropping it.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+          resourceClaims:
+          - name: gpu-claim
+            resourceClaimName: shared-trainers-gpu-claim
```
aishukamal, line 337: The referenced ResourceClaim is never defined in this guide and the key semantic of how to use it is unstated i.e. all jobs must reference the same claim object for sharing to happen. Suggest adding the claim manifest plus a one-line "one claim per group, shared by all jobs" note.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+Once the platform is deployed and the Slime code is integrated, you can submit multiple jobs and observe them sharing the GPUs.
+
+### Step 1: Submit Multiple Jobs
+Deploy two independent Slime jobs to the cluster (e.g., `slime-job-a` and `slime-job-b`).
+Ensure they have unique `TIMESLICE_JOB_ID` environment variables.
+
+### Step 2: Port-Forward the Orchestrator
+To monitor the orchestrator state from your local machine, port-forward the gRPC service:
+```bash
+kubectl port-forward svc/timeslice-acceleratororchestrator 50051:50051 -n timeslice-system
```
aishukamal, line 356: Service name check: `timeslice-acceleratororchestrator` here vs `acceleratororchestrator/accelerator-orchestrator` in the generic guide. Worth verifying both against actual helm output and fixing whichever doc is wrong.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+1. **Orchestrator Logs (Scheduling Decisions)**:
+   ```bash
+   kubectl logs -n timeslice-system -l app.kubernetes.io/name=acceleratororchestrator --tail=100 -f
+   ```
+   Look for lines indicating lock transfers:
+   ```text
+   [INFO] Acquire request from job "slime-job-b" for group "samplers" - Queued (Lock held by "slime-job-a")
+   [INFO] Yield received from job "slime-job-a" for group "samplers"
+   [INFO] Granting lock to next waiter "slime-job-b" for group "samplers"
```
aishukamal, line 386: Do these log excerpts match the orchestrator's actual slog output? How do we keep them in sync with the source of truth?

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+1. **Reward/Score Curves**: The mean reward should steadily increase over iterations, indicating the policy is successfully learning.
+2. **Policy & Value Loss**: Megatron-LM's training loss curves (actor loss, critic/value loss) should stabilize or decrease as training progresses.
+3. **KL Divergence**: Monitor the KL divergence between the active policy and the reference model to ensure it stays within target bounds (e.g., to prevent policy collapse).
+4. **Step vs. Wall-Clock Time**:
+   * **Step-wise Convergence**: The step-wise convergence graph (e.g., Reward vs. Training Steps) will align perfectly with a standalone (non-timesliced) run. The time-slicing process does not alter the mathematical state transitions.
+   * **Wall-Clock Progress**: Because the GPUs are shared, the wall-clock time per step will increase by a factor of $N$ (where $N$ is the number of co-located jobs), minus any gains from overlapping CPU-heavy phases (like reward processing or data loading) of one job with the other job's GPU phases.
```
aishukamal: "increase by a factor of N" is the absolute worst case (fully overlapping GPU phases between trainers and samplers) for which we may not even opt for time-slicing. For the pipelined setup of sync RL this effort targets, the increase should be well below N and that gap is the value proposition, so please reframe.

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
+# Time-Slicing Integration Guide for Slime Workloads
+
+This guide provides step-by-step instructions on how to integrate and deploy **Slime** (high-performance RL framework for LLMs) with the **llm-d-rl-time-slicing** platform.
+
+### Motivation: Maximizing GPU Utilization
+In traditional disaggregated RL setups, GPUs sit idle whenever worker groups wait for another phase to complete (e.g., trainer GPUs idling during rollout generation, or rollout GPUs idling during policy updates). Cooperative time-slicing enables multiple independent Slime jobs to multiplex physical GPU resource pools concurrently. When one job finishes a phase, its GPU context is checkpointed and evicted, allowing another job to immediately utilize the hardware—significantly driving up GPU duty cycle and overall cluster throughput.
+
+For a step-by-step reproduction guide using RayCluster Kubernetes manifests and disaggregated GRPO launch scripts, see:
+* **[Multi-Cluster Time-Sliced Disaggregated GRPO Guide](sync/README.md)**
```
aishukamal: This framing is a bit confusing: this README _is_ the step-by-step guide, while sync/ is a scripted example, but the link calls sync/ a "step-by-step reproduction guide" (and "Multi-Cluster" reads as multi-K8s-cluster rather than one Ray cluster per job). I suggest something like: "To see it working before integrating your own job, run the pre-scripted minimal runnable example: [Runnable example: two time-sliced GRPO jobs](sync/README.md)."

guides/rl-frameworks/slime/README.md
```diff
@@ -0,0 +1,385 @@
...
+> For a tested codebase reference branch containing these exact changes, see [jessicaochen/slime (timeslice branch)](https://github.com/jessicaochen/slime/tree/timeslice).
+> For a detailed walkthrough of all codebase changes made in this fork (categorized by scheduling, device leak fixes, and memory offloading), see **[Detailed Fork Changes](sync/SLIME_CHANGES.md)**.
+
+---
+
+## Concrete Deployment & Benchmark Reference
+
+For a step-by-step reproduction guide using RayCluster Kubernetes manifests, disaggregated GRPO launch scripts, and verified performance benchmark numbers, see:
+
+* **[Disaggregated Time-Sliced Slime Deployment Guide](sync/README.md)**
```
aishukamal: Same document is linked under a different name ("Disaggregated Time-Sliced Slime Deployment Guide"). I suggest reusing the exact title from L9 and retitling this section "Runnable Example & Benchmark" so readers see one guide + one example, not multiple overlapping guides.

guides/rl-frameworks/slime/sync/ray-job.yaml.template
```diff
@@ -0,0 +1,172 @@
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: ${JOB_NAME}
+spec:
+  entrypoint: /opt/scripts/run_disaggregated_grpo.sh
+  shutdownAfterJobFinishes: true
+  ttlSecondsAfterFinished: 60
+  runtimeEnvYAML: |
+    env_vars:
```
aishukamal, line 10: Are we missing a JOB_NAME env var here? `envsubst` only substitutes the YAML but the entrypoint script arrives via ConfigMap unsubstituted, so unless JOB_NAME is exported into the runtime (e.g. under `runtimeEnvYAML.env_vars`, where envsubst would catch it), the script falls back to slime-job-$RANDOM. That id drives the orchestrator lock, and if it doesn't match the pod labels, time-slicing breaks completely, no?

If the above is true, I'd suggest a full clean run of this recipe from just the files in this PR to make sure it works end to end. The plumbing gap above makes me think the tested setup and the shipped files may have drifted, so worth reconfirming.

guides/rl-frameworks/slime/sync/ray-job.yaml.template
```diff
@@ -0,0 +1,172 @@
...
+      PYTHONPATH: "/root/Megatron-LM"
+      CUDA_DEVICE_MAX_CONNECTIONS: "1"
+
+  submitterPodTemplate:
+    spec:
+      restartPolicy: Never
+      nodeSelector:
+        cloud.google.com/gke-nodepool: utility-pool-large
```
aishukamal: This is cluster-specific, no? Please remove.

guides/rl-frameworks/slime/sync/ray-job.yaml.template
```diff
@@ -0,0 +1,172 @@
...
+  rayClusterSpec:
+    rayVersion: '2.9.0'
+    enableInTreeAutoscaling: true
+    headGroupSpec:
+      rayStartParams:
+        dashboard-host: '0.0.0.0'
+      template:
+        spec:
+          nodeSelector:
+            cloud.google.com/gke-nodepool: utility-pool-large
```
aishukamal: Same here. This node selector should be removed.

guides/rl-frameworks/slime/sync/ray-job.yaml.template
```diff
@@ -0,0 +1,172 @@
...
+            slime-role: trainer
+            timeslice.io/job-id: "${JOB_NAME}"
+            timeslice.io/group: "trainers"
+        spec:
+          nodeSelector:
+            group.timeslice.io/trainers: "true"
+          tolerations:
+          - key: "nvidia.com/gpu"
+            operator: "Equal"
+            value: "present"
```
aishukamal: `value: "present"` is just GKE's auto-taint value string. Clusters that taint GPU nodes as `nvidia.com/gpu=true` or with no value won't match it. I suggest `operator: Exists` on the nvidia.com/gpu key to make it not GKE specific.

guides/rl-frameworks/slime/sync/ray-job.yaml.template
```diff
@@ -0,0 +1,172 @@
...
+            slime-role: rollout
+            timeslice.io/job-id: "${JOB_NAME}"
+            timeslice.io/group: "samplers"
+        spec:
+          nodeSelector:
+            group.timeslice.io/samplers: "true"
+          tolerations:
+          - key: "nvidia.com/gpu"
+            operator: "Equal"
+            value: "present"
```
aishukamal: I suggest `operator: Exists` on the nvidia.com/gpu key to make it not GKE specific.

guides/rl-frameworks/slime/sync/README.md
```diff
@@ -0,0 +1,74 @@
+# Multi-Cluster Time-Sliced Disaggregated GRPO Example
```
aishukamal, line 1: Really like the runnable example - it plays a nanoGPT-style role for time-slicing in my mind. I suggest opening with one line stating the scale and intent - minimal two-job time-slicing example sized to verify the integration end-to-end in minutes, not to measure performance.

guides/rl-frameworks/slime/sync/README.md
aishukamal, line 1: What do you think about hosting this under examples/slime/sync/ instead of putting it under guides/? It makes the guide vs runnable split obvious at the repo level.

guides/rl-frameworks/slime/sync/setup_node.sh
```diff
@@ -0,0 +1,27 @@
...
+
+echo "[Setup] Updating Slime codebase to timeslice branch..."
+git -C /root/slime remote add fork https://github.com/jessicaochen/slime.git || true
+git -C /root/slime fetch fork
+git -C /root/slime checkout timeslice
+git -C /root/slime reset --hard fork/timeslice
+
+echo "[Setup] Installing python client library..."
+pip install -U grpcio
+pip install git+https://github.com/llm-d-incubation/llm-d-rl-time-slicing.git#subdirectory=pkg/client/python
```
aishukamal: Can we pin this to a SHA or tag? We've been pushing breaking changes to the client recently and we definitely won't revalidate this recipe with every change, so let's pin to a working version.

guides/rl-frameworks/slime/README.md
jessicaochen: No, it is k8s be default now.

guides/rl-frameworks/slime/README.md
jessicaochen: Done, I aligned on 34 for ResourceClaims v1 in both guides.

guides/rl-frameworks/slime/README.md
jessicaochen: DRA relies on a plugin to advertise the GPU resources. That plugin itself has a requirement on the GPU driver version.

Removed that latter sentence about snapshot/restore which is misleading.

guides/rl-frameworks/slime/README.md
jessicaochen: Sure, added.

guides/rl-frameworks/slime/README.md
jessicaochen: Added

guides/rl-frameworks/slime/sync/ray-job.yaml.template
jessicaochen: Done

guides/rl-frameworks/slime/sync/ray-job.yaml.template
jessicaochen: Done

guides/rl-frameworks/slime/sync/README.md
jessicaochen: Done adding qualification that it is a simple quick example

guides/rl-frameworks/slime/sync/setup_node.sh
jessicaochen: Good point. Pinned.

guides/rl-frameworks/slime/sync/README.md
jessicaochen: I do not mind but want to do that in a different PR

guides/rl-frameworks/slime/sync/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+This directory contains the files required to deploy two independent sync RL jobs, each provisioned in its **own distinct Ray Cluster**, concurrently sharing physical GPU pools using Kubernetes **Dynamic Resource Allocation (DRA) Overloading** and the **Accelerator Orchestrator** time-slicing lock queues. This serves as a simple, quick example of the behavioral mechanisms of time slicing.
```
droot, line 3: Not blocker/critical for this PR but curious about pattern -- what sort of isolation cluster provide in Ray.

If the expectation is to not share resources with other jobs, then we might breaking that expectation with sharing the GPUs..

guides/rl-frameworks/slime/sync/README.md
```diff
@@ -0,0 +1,74 @@
...
+### Step 2.3: Submit the RayJobs Concurrently
+Using the parameterized template, submit both jobs concurrently using `envsubst` to set distinct job names:
+
+```bash
+# Submit Job 1
+JOB_NAME=slime-job1 envsubst < guides/rl-frameworks/slime/sync/ray-job.yaml.template | kubectl apply -f -
+
+# Submit Job 2
+JOB_NAME=slime-job2 envsubst < guides/rl-frameworks/slime/sync/ray-job.yaml.template | kubectl apply -f -
+```
```
droot, line 48: This is a good starting point.

I think we also might need another guide (and also manifests organization) from a perspective of how we want platform and ai teams to leverage this from multi-tenancy aspect where platform teams can apply the one time configuration and consumer of the platform (AI teams) applies just the workload configuration.

droot (commented):
LGTM to me overall, I have a few minor clarifications and suggestions.

guides/rl-frameworks/slime/sync/README.md
jessicaochen: From what I understand some customers use a Ray cluster per RL Job & some use a shared Ray cluster. I tried both variants for integration & the cluster-per-RL job works best with the current MVP timeslicing.

Based purely on how easy it was to integrate and the lack of bugs, it seems Ray cleanly tolerates their cluster not having full resources of the node. This seems to be true of other resources as well (other than GPU). We ask around though the Ray community though if we want a strategic confirmation.

guides/rl-frameworks/slime/sync/README.md
jessicaochen: Indeed, the goal here is a working example. Structure that supports decoupling time slicing awareness from RL job would be a bit different.

I definetly prefer to have a working guide in & improve on that.

guides/rl-frameworks/slime/sync/README.md
andrewsykim: >If the expectation is to not share resources with other jobs, then we might breaking that expectation with sharing the GPUs..

I would imagine that this is a platform decision based by whoever is operating Ray clusters. But the GPU sharing should be opaque from workload / developer perspective.

jessicaochen:
@coderabbitai ignore

guides/rl-frameworks/slime/README.md
jessicaochen: I can switch to accelerator here as timeslicing coordination itself has no accelerator dependencies.

However, this guide is indeed specific to GPUs. A bunch of stuff here does not have tested TPU equivalents yet (eg. what is the plugin for TPU DRA, what is the "cuda checkpoint" for TPUs). As a grounded guide, I do not want to force abstraction where it currently does not work.

guides/rl-frameworks/slime/README.md
jessicaochen: Removed confusing bits

guides/rl-frameworks/slime/README.md
jessicaochen: Removed, this is remant from when I had benmarks in my local repo.

guides/rl-frameworks/slime/README.md
jessicaochen: I am fine with it including the minimal steps needed to have a functional setup. However, functional is not the same as runnable which is what is under sync.

guides/rl-frameworks/slime/README.md
jessicaochen: done

guides/rl-frameworks/slime/README.md
jessicaochen: done

guides/rl-frameworks/slime/README.md
jessicaochen: Done

guides/rl-frameworks/slime/README.md
jessicaochen: Added but earlier

guides/rl-frameworks/slime/README.md
jessicaochen: Yes, I do not think we need to actually keep them in sync (nor do I think we have a reasonable mechanism to do so). Logs in general are fine for debugging but are not contracts. APIs are contracts.

I think this is generally true of watching history via logs for software in general.

guides/rl-frameworks/slime/README.md
jessicaochen: It actually depends on how you decided to deploy. I'll make it consistent though.

guides/rl-frameworks/slime/README.md
jessicaochen: reframed to focus more on the gains.

guides/rl-frameworks/slime/README.md
jessicaochen: I dropped this whole section so not relevant anymore

guides/rl-frameworks/slime/sync/ray-job.yaml.template
jessicaochen: It does work end to end. I ran it a bunch of times for benchmarking. 

That said, there are enough changes from the review that I will make sure to do a manual run before merging....

guides/rl-frameworks/slime/sync/ray-job.yaml.template
jessicaochen: Removed

guides/rl-frameworks/slime/sync/ray-job.yaml.template
jessicaochen: Removed

guides/rl-frameworks/slime/sync/ray-job.yaml.template
jessicaochen: Did manual run & this was in the fixes amongst other things from review changes.

droot (approved):

aishukamal (approved):
Thanks for working through the comments!

PR #90, opened 2026-07-10
[app-endpoint 1/6] snapshot-agent: refactor Backend interface and split server by mode

aishukamal:
## Summary

Refactor the Backend interface and split server handlers by deployment mode.

### Interface change
`Backend.Snapshot` and `Backend.Restore` now take a `backends.Request{JobID, Config *pb.BackendConfig}` instead of `[]string` PIDs. Each backend extracts what it needs from the request and validates its own inputs. `JobID` carries server-side per-job context to backends whose target is not in the wire config (e.g. a future backend that resolves its target by job); adding more such context later extends `Request` without touching the interface or the proto.

### Server split by mode
- **Standalone**: passes the caller-provided `BackendConfig` directly to the backend. No PID extraction, no server-side validation. Adding a new backend requires zero server changes.
- **K8s**: resolves PIDs via pod discovery, wraps in `BuildCudaConfig()`, caches PIDs for restore. Existing behavior unchanged.

### Files changed
- `backends/checkpoint.go` — adds `Request`, interface takes `Request`
- `backends/cuda-checkpoint.go` — extracts PIDs via `ExtractPIDStrings()`, validates both Snapshot and Restore, adds `BuildCudaConfig()` helper
- `backends/noop.go` — updated signature
- `server/server.go` — standalone path passes config through, k8s path unchanged
- `cuda_checkpoint_test.go` — uses `Request`, added NoPIDs/NilConfig cases
- `server_internal_test.go` — removed server-side standalone validation test (backend validates now)

Depends on #84.

## Test plan
- [x] All existing tests pass
- [x] New validation tests: NoPIDs and NilConfig
- [x] Build verified via Cloud Build

## Summary by CodeRabbit

* **New Features**
  * Snapshot and restore now accept backend-specific configuration in the request.
  * Standalone mode uses the provided CUDA PID targets/config.
  * Kubernetes-mode CUDA snapshots can discover and reuse process IDs automatically.
* **Bug Fixes**
  * Missing or invalid CUDA PID targets now fail with clearer PID-related errors.
  * Unsupported backend types and deployment modes are rejected consistently.
* **Tests**
  * Updated CUDA checkpoint and server RPC tests to follow the new request/config flow.
  * Added coverage for no-PIDs and asynchronous operation failures.

PR #97, opened 2026-07-11
[app-endpoint 2/6] snapshot-agent: add application-aware backend config protos

aishukamal:
## Summary

Adds proto definitions for application-aware backends. Part 3 of the app-aware backends series (stacked on #90).

- **`AppEndpointConfig`** (`app_endpoint` in the `BackendConfig` oneof): suspend/resume an application-aware workload through its HTTP API. `endpoints` targets the servers; the `App` enum (`APP_VLLM`, `APP_SGLANG`) selects the HTTP dialect. A new HTTP-capable application is an enum value, not a new message.
- **`SuspendMode`**: `OFFLOAD` (state preserved in host memory) or `DISCARD` (state dropped; the application re-provisions it). `UNSPECIFIED` defers to the application's default. For SGLang the effective mode is fixed by server launch flags, so the field is advisory there.
- **`tags`**: free-form region strings (`weights`, `kv_cache`, ...), validated by the application.
- `getSnapshotBackendType` resolves `app_endpoint` → `BackendAppEndpoint`; Go and Python client bindings regenerated.

A follow-up series adds `AppChannelConfig` for workloads that register over a workload channel instead of exposing HTTP (embedded engines, trainers).

No backend implementation yet — that lands in the next PR.

## Testing

`go build`, `go test ./pkg/snapshot-agent/...`, golangci-lint: clean. Exercised end to end by the integration suite in #100.


pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
```diff
@@ -38,9 +38,57 @@ message CudaBackendConfig {
   ProcessTarget explicit_target = 1;
 }
 
+// App identifies the HTTP API dialect of an application-aware workload.
+// It grows only when a new HTTP dialect ships; workloads reached through a
+// registered workload channel never need an entry here.
+enum App {
+  APP_UNSPECIFIED = 0;
+  // vLLM's dev-mode sleep API: POST /sleep?level=N, POST /wake_up?tags=...
+  // Requires VLLM_SERVER_DEV_MODE=1 and --enable-sleep-mode on the server.
```
droot, line 47: Does vllm's dev mode has any performance implications ? Any insight why it isn't available in production mode ?

pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
```diff
@@ -38,9 +38,57 @@ message CudaBackendConfig {
...
+  APP_VLLM = 1;
+  // SGLang's memory API: POST /release_memory_occupation and
+  // POST /resume_memory_occupation (JSON body). Requires
+  // --enable-memory-saver; whether suspended weights are preserved is fixed
+  // at server launch by --enable-weights-cpu-backup (see SuspendMode).
+  APP_SGLANG = 2;
```
droot, line 53: Supporting vllm/sglang natively is a good idea. 

One way to extend this will be to support a generic web-hook URLs such as `wake_up_url` and `sleep_url`.

pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
```diff
@@ -38,9 +38,57 @@ message CudaBackendConfig {
...
+  // HTTP endpoint(s) of the application server(s), e.g. "http://localhost:8000".
+  repeated string endpoints = 2;
+  // See SuspendMode. For APP_SGLANG this is advisory: the effective mode is
+  // fixed by the server's launch flags.
+  SuspendMode mode = 3;
+  // Region tags, e.g. "weights", "kv_cache". Free-form strings validated by
+  // the application; if empty, the application's full default region set is
+  // used. On Snapshot, tags select what to suspend (where the application
+  // supports it); on Restore, what to bring back.
+  repeated string tags = 4;
```
droot, line 85: I am little confused about why mode and tags both are needed ? wouldn't tag decide what to discard and what to restore ?

pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
aishukamal: No performance impact AFAICT — `VLLM_SERVER_DEV_MODE=1` only registers extra HTTP routes (/sleep, /wake_up, /is_sleeping), inference is untouched. The vLLM docs' stated caution is that these are administrative endpoints that "can disrupt service" and should stay on trusted networks, not be exposed to users ([docs](https://docs.vllm.ai/en/latest/features/sleep_mode/)). So the gating is about exposure of admin controls, not feature maturity AFAIU.

pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
aishukamal: I had thought about it. A plain URL only covers vLLM's dialect - SGLang takes its parameters in a JSON body with headers, which a URL can't carry.

Also, caller-built URLs mean every caller must know each engine's API details and update them if the engine changes. The App enum keeps callers high level at "which app + mode/tags" while SA owns the API details and application semantics in one place.

pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
aishukamal: Based on the engines and trainers I've researched (vLLM, SGLang, TRT-LLM, LMDeploy, DeepSpeed), they all express the same two capabilities: what state to act on, and what happens to it. Tags express the what - memory regions like weights, kv_cache, optimizer_state. Mode expresses the fate - preserved in host RAM (OFFLOAD) or dropped and re-provisioned by the app (DISCARD).

One thing to note is that not every engine supports every combination of tags and mode and those requests can be rejected with a clear error but these two fields cover the vast majority of the vocabulary of every popular engine/trainer I could find.

pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
droot: Ack.

pkg/snapshot-agent/api/v1alpha1/snapshot_agent.proto
droot: I see, got it. Thanks.

droot (approved):

PR #112, opened 2026-07-20
Publish component images and fix Helm chart defaults

aishukamal:
The Helm charts default to `ghcr.io/llm-d/*` images that have never been published, so `helm install` fails at image pull out of the box. The release workflow could not have produced them: it is unmodified repo-template scaffolding that builds the root `Dockerfile` (the `cmd/main.go` placeholder stub) as a single repo-named image, targeting an org this repo's token cannot push to.

Changes:
- **CI publishes both components**: `ci-release.yaml` builds `acceleratororchestrator` and `snapshot-agent` from their own Dockerfiles and pushes to `ghcr.io/llm-d-incubation/llm-d-rl-time-slicing/<component>`. The docker-build-and-push action gains a `dockerfile` input; the Trivy scan now scans the images actually pushed.
- **Release model**: `latest` is published on every merge to main; versioned tags are published via a manual `workflow_dispatch` run with a `version` input (versioned runs do not move `latest`). The template's tag-push/release triggers are removed — note that pushing a `v*` git tag alone no longer publishes images.
- **Images are linux/amd64 only**: the snapshot-agent is x86_64-bound (cuda-checkpoint binary + CGO build), and the platform is adopted as a unit, so the orchestrator matches rather than advertising an arch the stack can't deliver.
- **Hermetic snapshot-agent image**: the Dockerfile fetches cuda-checkpoint pinned to `NVIDIA/cuda-checkpoint@00d5cce` in the builder stage instead of requiring a pre-populated (gitignored) `bin/`, so it builds from a fresh clone. The chart's initContainer mechanism is unchanged.
- **Chart defaults match what CI publishes**: both `values.yaml` point at the new image path with `tag: latest` and `pullPolicy: Always` (explicit on the initContainer too, tied to the same value).
- **Docs**: published-image location noted in the deploy READMEs; fixed the snapshot-agent README's make target (`snapshot-agent-image-push`).

Both component Dockerfiles were build-verified from a clean checkout (no `bin/` preparation), and `helm lint`/`helm template` render the new image references on both subcharts and the parent chart.

Post-merge step (one-time): the first publish creates both ghcr packages **private** by default — an admin must set them to public (Package settings → Change visibility) for anonymous pulls / out-of-the-box `helm install` to work. Visibility persists for all future pushes.

## Summary by CodeRabbit

* **New Features**
  * Releases can now run automatically after merges to `main` or manually with a specified version.
  * CI publishes Accelerator Orchestrator and Snapshot Agent container images with configurable tags and build settings.
  * Snapshot Agent images now include the CUDA checkpoint utility.

* **Bug Fixes**
  * Deployment configuration now consistently uses the published container images and always retrieves the configured image version.

* **Documentation**
  * Added instructions for published image locations, tags, and updated image publishing commands.

### DIFF
--- a/deploy/snapshot-agent/README.md
+++ b/deploy/snapshot-agent/README.md
@@ -2,6 +2,8 @@
 
 This directory contains the Helm chart for deploying the Snapshot Agent DaemonSet in a Kubernetes cluster.
 
+Public images are published to `ghcr.io/llm-d-incubation/llm-d-rl-time-slicing/*` by CI: `latest` on every merge to main; versioned tags via a manual workflow run.
+
 ## Prerequisites
 
 *   A Kubernetes cluster with GPU nodes (NVIDIA).
@@ -134,7 +136,7 @@ We use the provided `Makefile` targets to build and push the container image.
     ```
 2.  Run the following make target from the repository root to build and push the image:
     ```bash
-    make image-push-snapshot-agent
+    make snapshot-agent-image-push
     ```
     This will build the image and push it to `your-custom-registry.com/your-project/snapshot-agent:dev-<hash>`.
 
--- a/.github/actions/docker-build-and-push/action.yml
+++ b/.github/actions/docker-build-and-push/action.yml
@@ -15,6 +15,10 @@ inputs:
     required: false
     description: Container registry
     default: ghcr.io/llm-d
+  dockerfile:
+    required: false
+    description: Path to the Dockerfile to build
+    default: Dockerfile
   platforms:
     required: false
     description: Target platforms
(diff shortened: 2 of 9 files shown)


jessicaochen:
A separate change but we should cut a 1.0.0 at some point.

jessicaochen (approved):
Excellent, we really needed this for ease of adoption. Thanks!

deploy/snapshot-agent/README.md
```diff
@@ -134,7 +136,7 @@ We use the provided `Makefile` targets to build and push the container image.
     ```
 2.  Run the following make target from the repository root to build and push the image:
     ```bash
-    make image-push-snapshot-agent
+    make snapshot-agent-image-push
```
jessicaochen, line 139: (sanity check)

If not already done, make sure the Make file is going to push to the correct location by default. I think it has the old path hardcoded last I looked at it.

aishukamal: Good catch. Will update in a separate PR after testing the whole flow e2e.

PR #113, opened 2026-07-21
snapshot-agent tests: pin engine images and add TEST_NODE override

aishukamal:
## Summary

Follow-up to #100 review:

- Pin the engine images for reproducible suite results: `vllm/vllm-openai:v0.25.1` and `lmsysorg/sglang:v0.5.15` (current releases) instead of `:latest`. Engine versions are now bumped deliberately.
- `TEST_NODE` env override for the harness's node selection: the default pick (first node with a free GPU **by requests**) cannot see workloads that occupy GPUs without requesting them (time-slicing experiments) and can land the suite on a busy node.

## Testing

Full suite rerun with the pinned images on a dedicated H100 node: **9/9 pass** — see the results comment below.

## Summary by CodeRabbit

* **Documentation**
  * Added instructions for pinning integration tests to a specific Kubernetes node with `TEST_NODE`.

* **Tests**
  * Added support for selecting a specific test node, while retaining automatic GPU-node selection by default.
  * Updated test runner configuration to pass the node selection option.
  * Pinned vLLM and SGLang test images to specific versioned, digest-verified releases for more consistent test execution.

### DIFF
--- a/tests/integration/snapshot-agent/README.md
+++ b/tests/integration/snapshot-agent/README.md
@@ -66,6 +66,11 @@ This builds from your working directory, so local modifications are included —
 --skip-cleanup       Leave the test-runner pod running for debugging
 ```
 
+Environment: `TEST_NODE=<node-name>` pins the suite to a specific node instead
+of the default pick (first node with a free GPU by requests). Use it when the
+cluster runs workloads that occupy GPUs without requesting them (time-slicing
+experiments), which the default pick cannot see.
+
 ## Exit code
 
 `go test`'s exit code (0 = all passed).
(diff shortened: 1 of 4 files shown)


aishukamal:
Full suite with the pinned images (`vllm-openai:v0.25.1`, `sglang:v0.5.15`) on a dedicated H100 node (via `TEST_NODE`): **9/9 pass**.

```
PASS TestK8s (212.6s):        CUDAWatcherDiscoveredPIDs, VLLMSleepWake, SGLangReleaseResume
PASS TestStandalone (178.9s): CUDACheckpointRestore, VLLMSleepWake, VLLMCompound,
                              VLLMSuspendDiscard, SGLangReleaseResume, SGLangCompound
ok   tests/integration/snapshot-agent   391.5s
```

Note for the record: one earlier run showed all four vLLM tests failing with `/sleep` returning 200 but freeing nothing (~23 GB resident). Not version-related — `latest` and `v0.25.1` are the same digest, and the failure didn't reproduce (two subsequent green runs). Looks like a transient in vLLM's sleep path; if it recurs, the right hardening is for the agent to verify `/is_sleeping` (or a VRAM drop) instead of trusting the 200.

lynnl0927 (dismissed):

lynnl0927 (approved):

PR #119, opened 2026-07-24
snapshot-agent tests: exercise the Helm chart and Makefile install paths

aishukamal:
## Summary

Pure test changes: both integration phases now run against a real install path — the way a user deploys the agent, not a test-only build. Stacked on #122 (AGENT_PORT fix, needed for the chart's port override) which stacks on #121 (`make standalone`); retarget as those merge.

- **standalone**: `run.sh` builds `make standalone` in the test-runner pod and `TestStandalone` runs the agent from those artifacts (`bin/snapshot-agent` + `bin/cuda-checkpoint`) on a plain Debian base with the driver mounts. The test-only agent image and `Dockerfile.snapshot-agent-runtime` are removed; `cloudbuild-image.yaml` now defaults to the official Dockerfile (and gains the `_DOCKERFILE` substitution).
- **k8s**: `run.sh` installs the official snapshot-agent chart (pinned to `TEST_NODE`, image and agent port overridden) and `TestK8s` runs its unchanged scenarios against the chart's DaemonSet — every backend is exercised through the chart exactly once.
- `run.sh` is the single entrypoint (`--phase both` = standalone + k8s, the default). `--build` has Cloud Build produce the agent image from the working directory, tagged `integ-<commit>` so reruns don't fight the node's image cache. Chart fixtures are installed before `go test` starts and fail fast with a chart-attributed error if the pod never becomes Ready. `TEST_NODE` is required for the k8s phase so the suite stays off other workloads on shared clusters.
- The tests README documents the contributor dev loop: one `run.sh --build --project <p>` command tests uncommitted working-directory changes end to end (images, charts, make build, Python client, and tests all come from the workspace).

## Testing

`--phase both` green on an H100 GKE node, image built by `--build` from the working directory:

```
--- PASS: TestK8s (210.51s)
    --- PASS: TestK8s/CUDAWatcherDiscoveredPIDs (27.34s)
    --- PASS: TestK8s/VLLMSleepWake (3.38s)
    --- PASS: TestK8s/SGLangReleaseResume (2.37s)
--- PASS: TestStandalone (183.01s)
    --- PASS: TestStandalone/CUDACheckpointRestore (2.33s)
    --- PASS: TestStandalone/VLLMSleepWake (3.39s)
    --- PASS: TestStandalone/VLLMCompound (4.63s)
    --- PASS: TestStandalone/VLLMSuspendDiscard (2.33s)
    --- PASS: TestStandalone/SGLangReleaseResume (2.38s)
[...]

PR #120, opened 2026-07-24
tests: shared integration harness and single launcher layout

aishukamal:
## Summary

Stacked on #119. Pure refactor — no behavior change.

- `tests/integration` becomes the home of the integration suites with the launcher at the top: `tests/integration/{README.md, run.sh, runner.yaml, harness/, snapshot-agent/}`. `run.sh` stays the single entrypoint (paths updated; the gate workflow's README pointer follows the move).
- New `tests/integration/harness` package: the generic pieces extracted from the snapshot-agent harness — in-cluster client setup, node selection (`TEST_NODE` / first-free-GPU pick), pod lifecycle (ready-wait, force-delete, exec with stdin), HTTP and VRAM helpers, and attach-by-label for chart-deployed pods. `snapshot-agent/` keeps only the agent specifics (agent deployment, engine specs, the agentctl.py call path) and composes the shared `Cluster`.

## Verification

Cloud Build green (build, `go vet -tags integration ./tests/...`, unit tests, lint). Pure refactor with no runtime change, so no live cluster run.

## Summary by CodeRabbit

* **Documentation**
  * Updated integration test documentation, repository layout guidance, and command examples.
  * Improved failure messages to reference the current documentation location.

* **Integration Testing**
  * Added a Kubernetes-based integration test runner and shared test utilities.
  * Expanded test execution to cover the full integration test suite.
  * Improved support for node and GPU selection, pod readiness, command execution, and resource verification.
  * Simplified snapshot-agent integration test setup through shared testing capabilities.


jessicaochen:
I suppose a nit is that it should probably be test not tests following https://github.com/golang-standards/project-layout/blob/master/test/README.md

jessicaochen:
so I know with what expectation to review this with, is the orchestrator integegraion supposed to run with the code in this PR or is it purely moving the agent harness to the shared location before making it work with both agent & orchestrator e2e?

tests/integration/run.sh
jessicaochen, line 45: Do I always have to specify this info? I prefer it just talk to whatever cluster my kubectl is talking to.

tests/integration/snapshot-agent/harness.go
```diff
@@ -109,128 +102,46 @@ func NewHarness(t *testing.T) *Harness {
...
 		if err != nil {
 			t.Fatalf("invalid CHART_AGENT_PORT %q: %v", p, err)
 		}
 		h.AgentPort = port
 	}
-	h.AgentIP = h.waitChartAgentReady(t)
+	// The chart runs the agent on hostNetwork, so the pod IP is the node IP —
+	// also how the orchestrator reaches agents.
+	h.AgentIP = h.WaitPodReadyByLabel(t, chartNamespace,
+		"app.kubernetes.io/name=snapshot-agent,app.kubernetes.io/instance=sa-chart-test",
```
jessicaochen, line 128: (sanity check) Is this "sa-chart-test" specfic to where you were running it or it this a label that your harness ensures is applied?

tests/integration/snapshot-agent/harness.go
```diff
@@ -109,128 +102,46 @@ func NewHarness(t *testing.T) *Harness {
...
-	return used, nil
-}
-
 func (h *Harness) deployAgent(t *testing.T) {
 	t.Helper()
-	h.deletePodAndWait(t, agentPodName)
+	h.DeletePodAndWait(t, agentPodName)
 
 	pod := agentPod(h.Node)
 	if _, err := h.Client.CoreV1().Pods(namespace).Create(context.Background(), pod, metav1.CreateOptions{}); err != nil {
```
jessicaochen, line 139: (sanity check)

why is this agent deployment code still here? I thought deployment was done via helm in the runner?

jessicaochen (changes requested):

tests/integration/run.sh
aishukamal: No, cluster/zone/project are already optional. If you omit them, run.sh skips `gcloud get-credentials` and uses whatever cluster your kubectl is talking to. `--project` is only required with `--build` (Cloud Build needs it for the image registry). Will clarify the usage string.

tests/integration/snapshot-agent/harness.go
aishukamal: That's the Helm release name (passed as `fullnameOverride` during chart install in `run.sh`), not a cluster-specific label. We do it to avoid colliding with any existing release on the cluster. `run.sh` installs and uninstalls it.

tests/integration/snapshot-agent/harness.go
aishukamal: Snapshot agent supports k8s mode and a standalone mode (for bare metal). The standalone mode has no chart and just uses artifacts produced by the Makefile. This code deploys the agent from `make standalone` artifacts (binary + cuda-checkpoint built in the runner pod by run.sh) into a privileged pod that simulates a bare-metal host. The k8s phase uses the official chart (installed by run.sh before go test starts).

I think long-term, we should just use a GCE VM to run the standalone tests, so we don't need this pod to "simulate" a bare-metal environment. I'll file an issue for this.

aishukamal:
> so I know with what expectation to review this with, is the orchestrator integegraion supposed to run with the code in this PR or is it purely moving the agent harness to the shared location before making it work with both agent & orchestrator e2e?

No this PR only extracts the shared harness and moves the launcher up to test/integration/. The orchestrator integration suite is in #129 (draft), which builds on this shared infrastructure.

> I suppose a nit is that it should probably be test not tests following https://github.com/golang-standards/project-layout/blob/master/test/README.md

Ack, I'll do this in a separate PR once this chain is merged to avoid having to update 4 other PRs in the queue.

aishukamal:
Integration suite rerun from the refactored layout on a dedicated H100 node: **9/9 pass**.

```
PASS TestK8s (414.4s):        CUDAWatcherDiscoveredPIDs, VLLMSleepWake, SGLangReleaseResume
PASS TestStandalone (162.2s): CUDACheckpointRestore, VLLMSleepWake, VLLMCompound,
                              VLLMSuspendDiscard, SGLangReleaseResume, SGLangCompound
ok   tests/integration/snapshot-agent   576.7s
```

The k8s phase ran against the official chart (installed by run.sh); the standalone phase built from `make standalone` artifacts. The shared harness extraction introduced no behavioral change.

tests/integration/snapshot-agent/harness.go
jessicaochen: I think I could use at minimum a comment (or better a method rename) to indicate this is for the bare metal host.

tests/integration/README.md
```diff
@@ -78,10 +77,11 @@ TEST_NODE=<gpu-node> ./tests/integration/snapshot-agent/run.sh \
...
-                     gcloud get-credentials with --cluster)
---cluster CLUSTER    GKE cluster name
---zone ZONE          GKE cluster zone
+--project PROJECT    GCP project (required with --build for image pushes;
+                     also used by gcloud get-credentials with --cluster)
+--cluster CLUSTER    GKE cluster name (optional; omit to use current
+                     kubectl context)
+--zone ZONE          GKE cluster zone (optional)
 --model MODEL        Model to load (default: Qwen/Qwen2.5-0.5B)
 --phase PHASE        "standalone", "k8s", or "both" (default)
```
jessicaochen, line 88: Either here or somewhere else, I think it is needed to explain how standalone works (ie. you are actually still using a cluster just deploying a pod as the "standalone host") because it is not straighforward.

jessicaochen (changes requested):

tests/integration/snapshot-agent/harness.go
aishukamal: Renamed method.

tests/integration/README.md
aishukamal: Done.

jessicaochen (approved):

PR #123, opened 2026-07-24
tests: move the orchestrator scenarios next to the integration suites

aishukamal:
## Summary

Moves the orchestrator scenario code from `test/e2e/acceleratororchestrator/` to `tests/integration/orchestrator/scenarios/` — a shared test library consumed by both the fakes tier (in-process, every PR) and the composed integration tier (#129).

Files moved (package rename only, content unchanged):
- `scenarios.go` — scenario drivers (`RunSingleRLJobScenario`, `RunQueuedRLJobsScenario`)
- `fake_rl_job.go` — scripted RL job actor driving Acquire/Yield
- `pod_factory.go` — workload pod templates
- `fakes_test.go` — fake agent store, fake scheduler, fake queue
- `scenarios_test.go` — fakes-tier test cases (test the orchestrator against fakes)

The `test/e2e/` directory is deleted. The `rlts test` subcommand is unchanged; its removal is tracked separately in #140.

Directory structure within `orchestrator/scenarios/` will be refined in a follow-up (see Jessica's review feedback on separating the test harness from the scenario library).

## Testing

- Cloud Build green (build + unit tests including moved scenario tests + lint).
- SA integration suite (9/9 on H100) unaffected — this PR touches no snapshot-agent code.

### DIFF
--- a/cmd/rlts/cmd/test_e2e.go
+++ b/cmd/rlts/cmd/test_e2e.go
@@ -27,7 +27,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 
 	pb "github.com/llm-d-incubation/llm-d-rl-time-slicing/pkg/accelerator-orchestrator/api/v1alpha1"
-	"github.com/llm-d-incubation/llm-d-rl-time-slicing/test/e2e/acceleratororchestrator"
+	"github.com/llm-d-incubation/llm-d-rl-time-slicing/tests/integration/orchestrator/scenarios"
 )
 
 var testCmd = &cobra.Command{
@@ -163,7 +163,7 @@ var orchestratorTestCmd = &cobra.Command{
 
 		// Scenario 1: Single RL Job
 		fmt.Println("--- Running Scenario: Single RL Job ---")
-		err = acceleratororchestrator.RunSingleRLJobScenario(ctx, clientset, client, cliLogger, samplerTemplateKey, trainerTemplateKey)
+		err = scenarios.RunSingleRLJobScenario(ctx, clientset, client, cliLogger, samplerTemplateKey, trainerTemplateKey)
 		scenario1Passed := err == nil
 		if !scenario1Passed {
 			fmt.Printf("[FAIL] Single RL Job Scenario failed: %v\n\n", err)
@@ -173,7 +173,7 @@ var orchestratorTestCmd = &cobra.Command{
 
 		// Scenario 2: Queued RL Jobs
 		fmt.Println("--- Running Scenario: Queued RL Jobs ---")
-		err = acceleratororchestrator.RunQueuedRLJobsScenario(ctx, clientset, client, cliLogger, samplerTemplateKey, trainerTemplateKey)
+		err = scenarios.RunQueuedRLJobsScenario(ctx, clientset, client, cliLogger, samplerTemplateKey, trainerTemplateKey)
 		scenario2Passed := err == nil
 		if !scenario2Passed {
 			fmt.Printf("[FAIL] Queued RL Jobs Scenario failed: %v\n\n", err)
@@ -220,7 +220,7 @@ func init() {
 		"Name of the Kubernetes PodTemplate to use for trainer pods (blank for default pause pod)")
 }
 
-// cliLogger implements acceleratororchestrator.Logger interface to print to stdout.
+// cliLogger implements scenarios.Logger interface to print to stdout.
 type cliLogger struct{}
 
 func (c *cliLogger) Log(args ...interface{}) {
--- a/tests/integration/orchestrator/scenarios/fake_rl_job.go
+++ b/tests/integration/orchestrator/scenarios/fake_rl_job.go
@@ -1,4 +1,4 @@
-package acceleratororchestrator
+package scenarios
 
 import (
 	"context"
--- a/tests/integration/orchestrator/scenarios/pod_factory.go
+++ b/tests/integration/orchestrator/scenarios/pod_factory.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package acceleratororchestrator
+package scenarios
 
 import (
 	"sync"
--- a/tests/integration/orchestrator/scenarios/scenarios.go
+++ b/tests/integration/orchestrator/scenarios/scenarios.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package acceleratororchestrator
+package scenarios
 
 import (
 	"context"
--- a/tests/integration/orchestrator/simulate/fakes.go
+++ b/tests/integration/orchestrator/simulate/fakes.go
@@ -1,4 +1,4 @@
-package acceleratororchestrator
+package simulate
 
 import (
 	"context"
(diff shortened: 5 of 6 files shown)


pkg/accelerator-orchestrator/scenarios/pod_factory.go
```diff
@@ -125,13 +139,15 @@ func (p *PodFactory) Register(key string, pod *corev1.Pod) {
...
 // GetPod returns a copy of the registered pod template for the key.
-// If key is blank or not found, it falls back to the "default" template.
+// A blank key maps to the "pause" template (no GPU work — safe on clusters
+// whose GPUs must stay untouched, and what the rlts flags document); an
+// unknown key falls back to the "default" template.
 func (p *PodFactory) GetPod(key string) *corev1.Pod {
 	p.mu.RLock()
 	defer p.mu.RUnlock()
 
 	if key == "" {
```
jessicaochen, line 133: I disagree making pause template the default at this point. Because it will not exercise snapshot/restore when on a real cluster. Actually, it should actually just hang because the agent will not transition a pod with no GPU work to running. Have you tried running the test command with this change?

Please update the documentation if it is inconsistent.

jessicaochen:
Here or in a direct message, it would help a lot if I know what was your planned end state with your collection of changes. 

For example, I do not mind if pkg is where all the e2e/integration testing goes if rlts is the e2e/integration is the command for all said integration tests. However, I do think they should be together so one can easily find all the e2e/integration stuff they can use to conformance test their setup.

jessicaochen (changes requested):

pkg/accelerator-orchestrator/scenarios/pod_factory.go
aishukamal: This code is now obsolete.

aishukamal:
> Here or in a direct message, it would help a lot if I know what was your planned end state with your collection of changes.
> 
> For example, I do not mind if pkg is where all the e2e/integration testing goes if rlts is the e2e/integration is the command for all said integration tests. However, I do think they should be together so one can easily find all the e2e/integration stuff they can use to conformance test their setup.

We discussed this offline. The agreed upon end state is:
* `scenarios` become the single definition of the orchestrator tests and they run in two places automatically - against fakes in regular CI on every PR (`scenarios_test.go`) and against a real chart-deployed cluster in the gated integration suite (`orchestrator_test.go`). Not all scenarios run in both suites - for e.g. orchestrator-only tests only run in `scenarios_test.go` since it isn't really an "integration" test.
* Adding a new scenario makes automatic registry with either or both suites (fakes and real) straightforward.
* rlts remains a pure debugging/ops CLI (inspect, status, acquire/yield). Its `test` subcommand will be dropped because the integration suite already provisions infra and tests from source, so another launcher for the same scenarios is unnecessary.

So that boils down to the following dir structure:
* `test/integration/orchestrator/scenarios` contains `scenarios.go`, `fake_rl_job.go`, `scenarios_test.go`, `fakes_test.go`.
* `test/integration/orchestrator` contains `orchestrator_test.go` (which triggers the orchestrator+agent composition tests)
* (existing) `test/integration/snapshot-agent` contains `standalone_test.go` and `k8s_test.go` for snapshot-agent
* (PR 120) `test/integration/` contains `harness/`, `run.sh`, `runner.yaml`, `README.md` which are the common integration framework pieces used by both orchestrator and snapshot-agent.

tests/integration/orchestrator/scenarios/scenarios_test.go
jessicaochen, line 31: I believe we agreed that having the unit test harness for the scenarios should not be in the same folder as scenarios to avoid confusion

cmd/rlts/cmd/test_e2e.go
```diff
@@ -1,246 +0,0 @@
...
-		clientset, err := kubernetes.NewForConfig(config)
-		if err != nil {
-			return fmt.Errorf("failed to create kubernetes client: %w", err)
-		}
-
-		ctx := context.Background()
-
-		// 3. STEP 1: General Prerequisite Check (Verify Deployment)
-		fmt.Println("=== Step 1: Running Prerequisite Verification ===")
-		overallPassed := verifyCluster(ctx, clientset)
```
jessicaochen, line 72: Cluster verification & pre cleanup are in this command but seemingly are deleted with it. Can you make sure the logic for verification & cleanup is not lost for the version that runs against a real cluster?

jessicaochen:
I suppose a nit is that it should probably be test not tests following https://github.com/golang-standards/project-layout/blob/master/test/README.md

jessicaochen (changes requested):

cmd/rlts/cmd/test_e2e.go
aishukamal: `verifyCluster` does the following things and here's where each one lives now:

1. Orchestrator deployment ready -> run.sh's orchestrator phase installs the chart and calls `wait_chart_pods_ready`, which polls for the Ready condition with a timeout.
2. Snapshot agent daemonSet ready -> same pattern in run.sh's k8s phase: chart install + `wait_chart_pods_ready`.
3. Node labels exist -> run.sh creates the labels itself rather than checking for pre-existing ones; cleanup removes only the labels it added.

Note that the `verifyCluster` function itself still lives in inspect.go and `rlts inspect` still works for checking any deployment.

Cleanup:
1. Lock pre-clean (defensive Yields) is in the composed suite's harness in #129.
2. All infra cleanup happens in run.sh's `cleanup()`

tests/integration/orchestrator/scenarios/scenarios_test.go
aishukamal: Ack. How about moving scenarios_test.go to pkg/accelerator-orchestrator/orchestrator_scenarios_test.go (they test the orchestrator, not the scenarios library) and renaming fakes_test.go to fakes.go (the whole test/ tree is test-only so the _test.go exclusion isn't needed)? After this, scenarios/ is a pure testing helpers library and the actual test files live in the component source code directory.

aishukamal:
> I suppose a nit is that it should probably be test not tests following https://github.com/golang-standards/project-layout/blob/master/test/README.md

Ack, will do in a separate PR to avoid churn to the in-flight PRs.

aishukamal:
Verification that the scenario move broke nothing:

- **Cloud Build** (build + unit tests incl. moved orchestrator scenario tests + lint): green.
- **SA integration suite** (9/9 on H100): unaffected — this PR touches no snapshot-agent code.
- Orchestrator scenario tests (`TestSingleRLJobScenario`, `TestQueuedRLJobsScenario`) run from their new location in `go test ./...` and pass.

tests/integration/orchestrator/scenarios/scenarios_test.go
jessicaochen: This seems like it is getting messy mixing test with prod code if we have stuff under pkg (unit test harness) take dependency on the test directory. How about:

test/integration/orchestrator/scenarios # the common scenario logic
test/integration/orchestrator # the harness against a real cluster 
test/integration/orchestrator/simluate # the unit test harness that is essentially simulating a cluster with fakes and mocks

For stuff under test directory, I have no problem with things not using the *_test.go to hide the test logic.

cmd/rlts/cmd/test_e2e.go
jessicaochen: Thanks for letting me know where it all went. I will reqest that this change of deleting the existing command only get merged at the same time or serially with the PR(s) that will provide the runnable equivalent command on the new framework.

I do regularly run orchestrater integration testing with my changes to ensure I have not broken anything.

jessicaochen (changes requested):

cmd/rlts/cmd/test_e2e.go
aishukamal: Deferring the deletion. Tracked in #140.

tests/integration/orchestrator/scenarios/scenarios_test.go
aishukamal: Done. The new structure is:

```
test/integration/orchestrator/
  scenarios/                              ← shared building blocks (both tiers consume)
    scenarios.go                          ← scenario drivers: RunSingleRLJobScenario, RunQueuedRLJobsScenario
    fake_rl_job.go                        ← scripted RL job actor: drives Acquire/Yield against any orchestrator
    pod_factory.go                        ← workload pod templates registered by name

  simulate/                               ← fakes tier: orchestrator logic tested against a simulated cluster
    fakes.go                              ← was: fakes_test.go → renamed (no _test.go needed under test/)
                                            contents: FakeSnapshotAgentStore, StartFakeScheduler, trackQueue
    orchestrator_simulated_test.go        ← was: scenarios_test.go → renamed + moved
                                            contents: TestSingleRLJobScenario, TestQueuedRLJobsScenario
                                            (wires fakes + scenarios, runs every PR, no build tag)

  orchestrator_test.go                    ← NEW (#129): composed tier, real cluster, build-tagged
  harness.go                              ← NEW (#129): composed fixtures + assertions, build-tagged
```

jessicaochen (approved):

PR #129, opened 2026-07-24
tests: composed orchestrator integration suite against both official charts

aishukamal:
## Summary

Adds the composed orchestrator integration suite: real orchestrator scenarios running against BOTH official Helm charts (snapshot-agent + timeslice-orchestrator) deployed on a real cluster with real GPU workloads.

- `--phase orchestrator` installs the snapshot-agent chart (DaemonSet) on `CHART_AGENT_PORT` (default 9002) and the timeslice-orchestrator chart with `snapshotAgentPort` set to the same port, so the orchestrator commands the suite's own agent — not whatever might be running on the default port on a shared node.
- Canonical 2-node topology: `TEST_NODE_SAMPLERS` and `TEST_NODE_TRAINERS` (one GPU per group, separate nodes). `exclusiveLabel` temporarily removes group labels from other nodes so pods land only on the designated test nodes.
- `TestOrchestrator/SingleRLJob`: single job acquires locks, deploys GPU pods, runs 2 sampling/training iterations, cleans up.
- `TestOrchestrator/QueuedRLJobs`: two jobs contend for the same locks — Job B queues behind Job A, both complete 2 iterations.
- The orchestrator chart gains a `snapshotAgentPort` value passed as `--snapshot-agent-port` to the binary.
- Runner RBAC gains node update/patch (for group labeling) and `resourceclaims` create/delete/get/list (for the scenarios' DRA claims).

### Files added

| File | Purpose |
|------|---------|
| `tests/integration/orchestrator/orchestrator_test.go` | `TestOrchestrator` with `SingleRLJob` and `QueuedRLJobs` subtests |
| `tests/integration/orchestrator/harness.go` | `ComposedHarness`: 2-node topology, attaches to chart-deployed orchestrator + agent, exclusive group labeling, pre-cleans leaked pods |

### Files modified

| File | Change |
|------|--------|
| `tests/integration/run.sh` | `--orch-image` flag, `--phase orchestrator` and `all`, 2-node topology (`TEST_NODE_SAMPLERS`/`TEST_NODE_TRAINERS`), orchestrator chart install/cleanup |
| `tests/integration/runner.yaml` | Node update/patch RBAC, ResourceClaims RBAC |
| `tests/integration/README.md` | Orchestrator phase documentation |
| `tests/integration/harness/harness.go` | `WaitPodReadyByLabel` handles empty node |
| `deploy/timesliceorchestrator/values.yaml` | `snapshotAgentPort` value |
| `deploy/timesliceorchestrator/templates/deployment.yaml` | Pass `--snapshot-agent-port` arg |
| `tests/integration/orchestrator/scenarios/fake_rl_job.go` | Force-delete pods on cleanup (`GracePeriodSeconds=0`) |
| `tests/integration/orchestrator/scenarios/scenarios.go` | Increased post-cleanup verification timeout (500ms/10s) |

### Integration test results

2-node L4 topology, PyTorch GPU burner template:

| Test | Result | Duration |
|------|--------|----------|
| `TestOrchestrator/SingleRLJob` | PASS | 25.52s |
| `TestOrchestrator/QueuedRLJobs` | PASS | 115.24s |
| **TestOrchestrator (total)** | **PASS** | **141.52s** |

Cloud Build verification (build + `go vet -tags integration` + unit tests): **SUCCESS**

### DIFF
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -14,9 +14,9 @@ Tests snapshot-agent backends in standalone and k8s modes. The Go harness deploy
 
 ### orchestrator (phase: orchestrator)
 
-Composed orchestrator integration suite. Installs BOTH official Helm charts (snapshot-agent + timeslice-orchestrator) on `TEST_NODE` and drives real orchestrator scenarios through the gRPC API. The orchestrator chart is configured with `snapshotAgentPort` matching `CHART_AGENT_PORT` so it commands the suite's own agent.
+Composed orchestrator integration suite. Installs BOTH official Helm charts (snapshot-agent + timeslice-orchestrator) and drives real orchestrator scenarios through the gRPC API. The orchestrator chart is configured with `snapshotAgentPort` matching `CHART_AGENT_PORT` so it commands the suite's own agent.
 
-The suite labels `TEST_NODE` with dedicated integration groups (`integ-samplers`, `integ-trainers`) so scenario workloads do not interfere with unrelated production groups on shared clusters.
+Uses a 2-node topology: `TEST_NODE_SAMPLERS` for the samplers group, `TEST_NODE_TRAINERS` for the trainers group (one GPU per node, separate nodes). `exclusiveLabel` temporarily removes group labels from other nodes so pods land only on the designated test nodes.
 
 ## Layout
 
@@ -70,12 +70,14 @@ so no commit or merge is needed at any layer:
 TEST_NODE=<gpu-node> ./tests/integration/run.sh \
   --build --project <gcp-project>
 
-# orchestrator phase only:
-TEST_NODE=<gpu-node> ./tests/integration/run.sh \
+# orchestrator phase only (2 GPU nodes, one per group):
+TEST_NODE_SAMPLERS=<gpu-node-1> TEST_NODE_TRAINERS=<gpu-node-2> \
+  ./tests/integration/run.sh \
   --build --project <gcp-project> --phase orchestrator
 
-# everything:
-TEST_NODE=<gpu-node> ./tests/integration/run.sh \
+# everything (standalone + k8s on TEST_NODE, orchestrator on the 2-node topology):
+TEST_NODE=<gpu-node> TEST_NODE_SAMPLERS=<gpu-node-1> TEST_NODE_TRAINERS=<gpu-node-2> \
+  ./tests/integration/run.sh \
   --build --project <gcp-project> --phase all
 ```
 
@@ -86,7 +88,8 @@ caches).
 Alternative (pre-built images -- any registry the cluster can pull from):
 
 ```bash
-TEST_NODE=<gpu-node> ./tests/integration/run.sh \
+TEST_NODE=<gpu-node> TEST_NODE_SAMPLERS=<gpu-node-1> TEST_NODE_TRAINERS=<gpu-node-2> \
+  ./tests/integration/run.sh \
   --agent-image gcr.io/<project>/snapshot-agent:dev \
   --orch-image gcr.io/<project>/timesliceorchestrator:dev \
   --phase all
@@ -116,11 +119,15 @@ TEST_NODE=<gpu-node> ./tests/integration/run.sh \
 
 Environment:
 
-- `TEST_NODE=<node-name>` -- required for the k8s and orchestrator phases
+- `TEST_NODE=<node-name>` -- required for the standalone and k8s phases
   (charts are pinned to this node); for the standalone phase it pins the
   suite instead of the default pick (first node with a free GPU). Use the
   pin when the cluster runs workloads that occupy GPUs without requesting
   them (time-slicing experiments), which the default pick cannot see.
+- `TEST_NODE_SAMPLERS=<node-name>` -- required for the orchestrator phase.
+  GPU node for the samplers group (must be different from `TEST_NODE_TRAINERS`).
+- `TEST_NODE_TRAINERS=<node-name>` -- required for the orchestrator phase.
+  GPU node for the trainers group (must be different from `TEST_NODE_SAMPLERS`).
 - `CHART_AGENT_PORT=<port>` -- port for the chart-deployed agent (default
   9002), so the suite can coexist with an unrelated agent on the default
   port (the chart runs on hostNetwork). The orchestrator chart is
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -1,21 +1,36 @@
 # Integration Tests
 
-End-to-end tests for snapshot-agent backends on real GPU hardware, in k8s and standalone modes, exercising the Helm chart and Makefile deployment paths respectively.
+End-to-end tests for the time-slicing stack on real GPU hardware, exercising the official Helm chart deployment paths.
 
-The test suite is written in Go and runs inside the cluster: `run.sh` deploys a test-runner pod, copies the repo source into it, and executes `go test` there. The Go harness deploys the snapshot-agent and inference engine pods itself — one engine at a time, so a single free GPU is enough.
+The test suite is written in Go and runs inside the cluster: `run.sh` deploys a test-runner pod, copies the repo source into it, and executes `go test` there.
 
-All snapshot/restore calls go through the **Python client** (`timeslice.snapshot_agent`, invoked via `agentctl.py`), so the entire client layer is covered.
+## Suites
 
-- `run.sh` — launcher (build image, install chart fixture, deploy runner, copy source, build `make standalone`, install the Python client, `go test`, cleanup)
-- `runner.yaml` — test-runner pod + RBAC
-- `harness/` — shared framework: in-cluster client, node selection, pod lifecycle, exec/HTTP/VRAM helpers
-- `snapshot-agent/` — the agent suite: `standalone_test.go` / `k8s_test.go`, plus the agent specifics (`harness.go` agent deployment, `engines.go` engine specs, `agentctl.py` — a thin CLI over the Python client that builds `BackendConfig` protos from primitive flags)
+### snapshot-agent (phases: standalone, k8s)
 
-**How the standalone mode for snapshot-agent works:** since the test suite runs inside a GKE cluster, standalone mode is simulated by deploying a privileged pod with `hostPID` and `hostNetwork` on the test node. The `make standalone` artifacts are built in the test runner and copied into this pod, which then runs the agent binary with the same GPU and PID namespace access as a host process. Long-term, standalone tests will run on an actual GPU VM.
+Tests snapshot-agent backends in standalone and k8s modes. The Go harness deploys the snapshot-agent and inference engine pods itself -- one engine at a time, so a single free GPU is enough. All snapshot/restore calls go through the **Python client** (`timeslice.snapshot_agent`, invoked via `agentctl.py`), so the entire client layer is covered.
+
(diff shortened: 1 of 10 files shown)


aishukamal:
Composed orchestrator suite on 2×L4 GPU node: **PASS**

```
PASS TestOrchestrator/SingleRLJob (416.8s)
ok   tests/integration/orchestrator   417.4s
```

Both official charts deployed (SA on port 9002, orchestrator with `snapshotAgentPort=9002`). The scenario ran the full SingleRLJob lifecycle: init (deploy sampler + trainer pods), 2 loop iterations (acquire/yield each group), cleanup. Labels temporarily removed from the production H100 node and restored on cleanup.

QueuedRLJobs excluded (blocked by the orchestrator cold-start IDLE bug).

tests/integration/orchestrator/harness.go
```diff
@@ -0,0 +1,210 @@
...
+		agentPort = port
+	}
+	t.Logf("snapshot-agent chart pod ready at %s:%d on %s", agentIP, agentPort, node)
+
+	// Ensure TEST_NODE is the ONLY node with these group labels so the
+	// scheduler pins scenario pods there. Other nodes (e.g. a production
+	// timeslice release) may carry the same labels; temporarily remove
+	// them for the test and restore on cleanup.
+	for _, group := range []string{integSamplers, integTrainers} {
+		h.exclusiveLabel(t, node, group)
```
jessicaochen: I could be crazy but are you putting both the sampler and the trainers on the exact same node?

I am pretty sure you do not want different groups on the same node in our normal scenario we want to test in a real k8s cluster. That ability is for sub-node which we have not gotten around to supporting (ie. pretty sure there will be bugs from trying this).

tests/integration/orchestrator/orchestrator_test.go
```diff
@@ -0,0 +1,65 @@
...
+		defer conn.Close()
+
+		client := pb.NewTimeSliceOrchestratorServiceClient(conn)
+
+		err = scenarios.RunSingleRLJobScenario(
+			ctx,
+			h.Client,
+			client,
+			t,
+			"vllm",  // sampler template key
```
jessicaochen: We should default to that pod template that runs pytourch for speed and simplicity. 

vllm can be an option but it should not be the default. Perhaps we can make this configurable like it was for the cli command?

tests/integration/README.md
```diff
@@ -1,21 +1,36 @@
...
+
+**How the standalone mode works:** since the test suite runs inside a GKE cluster, standalone mode is simulated by deploying a privileged pod with `hostPID` and `hostNetwork` on the test node. The `make standalone` artifacts are built in the test runner and copied into this pod, which then runs the agent binary with the same GPU and PID namespace access as a host process.
+
+### orchestrator (phase: orchestrator)
+
+Composed orchestrator integration suite. Installs BOTH official Helm charts (snapshot-agent + timeslice-orchestrator) on `TEST_NODE` and drives real orchestrator scenarios through the gRPC API. The orchestrator chart is configured with `snapshotAgentPort` matching `CHART_AGENT_PORT` so it commands the suite's own agent.
+
+The suite labels `TEST_NODE` with dedicated integration groups (`integ-samplers`, `integ-trainers`) so scenario workloads do not interfere with unrelated production groups on shared clusters.
```
jessicaochen: similar to the other comment, for an actual test orchestrator & agent we should be using 2 gpu nodes. one for trainer and one for samplers. 

This sounds a like it is putting both groups into one machine which is not the canonical setup we are saying customers should do.

jessicaochen:
> QueuedRLJobs

Let's not check this in till QueuedRLJobs works. Do not want to lose scenario e2e coverage. One interesting question is why QueuedRLJobs works on the CLI but not in this harness setup? it is using the same scenario logic which should be deploying the RL GPU pod that will come to ready under lock. What is different here in this harness setup?

jessicaochen (changes requested):

tests/integration/orchestrator/scenarios/fake_rl_job.go
```diff
@@ -418,9 +418,10 @@ func (f *FakeRLJob) cleanupPods(ctx context.Context) error {
 	f.mu.Lock()
 	defer f.mu.Unlock()
 
-	// 1. Delete Pods
+	// 1. Delete Pods (force-delete so they vanish immediately)
+	zero := int64(0)
 	for _, podName := range f.createdPods {
-		err := f.clientset.CoreV1().Pods("default").Delete(ctx, podName, metav1.DeleteOptions{})
+		err := f.clientset.CoreV1().Pods("default").Delete(ctx, podName, metav1.DeleteOptions{GracePeriodSeconds: &zero})
```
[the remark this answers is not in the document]
aishukamal: This is test-only code. Force deletion is preferred over polling and waiting which increases test time.

tests/integration/orchestrator/orchestrator_test.go
```diff
@@ -0,0 +1,89 @@
...
+			h.Client,
+			client,
+			t,
+			"",  // sampler template key (default: PyTorch GPU burner)
+			"",  // trainer template key (default: PyTorch GPU burner)
+		)
+		if err != nil {
+			t.Fatalf("QueuedRLJobs scenario failed: %v", err)
+		}
+	})
```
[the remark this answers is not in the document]
aishukamal: QueuedRLJobs is not impacted by the cold-start bug. The cold-start bug only impacts tests which deploy pods first and then acquire the lock which this test suite does not do.

tests/integration/orchestrator/harness.go
aishukamal: Yes, that's what the code was doing and only because I had only a 2 GPU node available for testing (and my agent picked up on that) and it actually worked fine. That said, I see your point - using separate nodes is a more accurate representation of the production use case, switched to a 2-node topology.

tests/integration/README.md
aishukamal: Done, switched to the 2-node setup. Updated README.

tests/integration/orchestrator/orchestrator_test.go
aishukamal: Done. Default template is now PyTorch GPU burner. vLLM is still available as a registered template for explicit use.

aishukamal:
> Let's not check this in till QueuedRLJobs works. Do not want to lose scenario e2e coverage. One interesting question is why QueuedRLJobs works on the CLI but not in this harness setup? it is using the same scenario logic which should be deploying the RL GPU pod that will come to ready under lock. What is different here in this harness setup?

QueuedRLJobs is now enabled and passing. The original exclusion was a misdiagnosis — FakeRLJob uses acquire-before-deploy (Acquire → deploy pod → wait ready), so the pod doesn't exist when Acquire is called. The agent reports UNSPECIFIED, which `isJobLoaded` treats as grantable. This is the same model the CLI uses, which is why it works there too.

The #137 cold-start bug only affects the deploy-before-acquire path (e.g. Ray clusters where worker pods exist before the job driver calls Acquire). Neither SingleRLJob nor QueuedRLJobs hits that path. We actually have a TODO on #137 to add e2e scenarios for that path whenever we fix the bug.

tests/integration/orchestrator/harness.go
```diff
@@ -0,0 +1,223 @@
...
+		if err != nil {
+			t.Logf("WARNING: listing leaked pods for group %s: %v", group, err)
+			continue
+		}
+		for i := range pods.Items {
+			name := pods.Items[i].Name
+			if err := h.Client.CoreV1().Pods("default").Delete(ctx, name, metav1.DeleteOptions{}); err != nil {
+				t.Logf("WARNING: deleting leaked pod %s: %v", name, err)
+			} else {
+				t.Logf("cleaned leaked pod %s", name)
```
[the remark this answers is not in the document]
aishukamal: Low risk in practice. The test nodes are exclusively labeled by the harness, so only test pods should be on those nodes with those labels. Hardening this logic won't hurt but deferrable.

tests/integration/runner.yaml
```diff
@@ -28,16 +28,17 @@ metadata:
...
+# namespace-local below.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: snapshot-agent-test-runner
 rules:
   - apiGroups: [""]
     resources: ["nodes"]
-    verbs: ["get", "list"]
+    verbs: ["get", "list", "update", "patch"]
```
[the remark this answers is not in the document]
aishukamal: Not actionable — Kubernetes RBAC can't restrict by resource name for node labels (no resourceNames scoping for label mutations). The test runner already runs on a dedicated test cluster.

tests/integration/orchestrator/harness.go
```diff
@@ -0,0 +1,210 @@
...
+				t.Logf("WARNING: could not restore label %s on %s: %v", labelKey, nodeName, err)
+				return
+			}
+			nn.Labels[labelKey] = "true"
+			if _, err := h.Client.CoreV1().Nodes().Update(ctx, nn, metav1.UpdateOptions{}); err != nil {
+				t.Logf("WARNING: could not restore label %s on %s: %v", labelKey, nodeName, err)
+			} else {
+				t.Logf("restored label %s on node %s", labelKey, nodeName)
+			}
+		})
```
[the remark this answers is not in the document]
aishukamal: Already addressed in the current code — `exclusiveLabel` records which nodes had labels removed and restores them in `t.Cleanup`. The code also checks if `n.Labels[labelKey] != "true"` before removing, and restores only what was removed. The concern about runner crash leaving labels orphaned is real but inherent to any test that mutates cluster state — `run.sh` cleanup is the backstop. Not worth changing further — the current implementation is correct.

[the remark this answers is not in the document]
aishukamal: The test nodes (TEST_NODE_SAMPLERS / TEST_NODE_TRAINERS) are dedicated integration test nodes — they don't have pre-existing group labels. The operator explicitly chooses which nodes to use for the test by setting the env vars. If the nodes already had the correct labels, labelNode writes the same value (true) and cleanup removes it, which is the correct behavior — the test should leave nodes clean, not preserve labels it didn't set up.

For run.sh cleanup: it's a last-resort backstop for the case where the Go process crashes. Its job is to remove test artifacts, not to be a full state-restoration system. The t.Cleanup path in Go handles the normal case correctly (including restoring labels on other nodes).

Adding label-state recording adds complexity for a scenario that doesn't exist in practice (running integration tests on nodes that are simultaneously serving production time-slicing groups). If someone does that, they have bigger problems than label preservation.

tests/integration/orchestrator/harness.go
```diff
@@ -0,0 +1,223 @@
...
+
+	n, err := h.Client.CoreV1().Nodes().Get(ctx, node, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("getting node %s: %v", node, err)
+	}
+	if n.Labels == nil {
+		n.Labels = make(map[string]string)
+	}
+	n.Labels[labelKey] = "true"
+	if _, err := h.Client.CoreV1().Nodes().Update(ctx, n, metav1.UpdateOptions{}); err != nil {
```
jessicaochen, line 181: (nit) You will want retry for update type operations in k8s in general. They can be sensitive to timing otherwise.

jessicaochen: True for all the other update calls you have introduced

jessicaochen (approved):

tests/integration/orchestrator/harness.go
aishukamal: I've got a list of test framework robustness improvements tracked in #143. Added this to that list and will address them together.

Issue #134, opened 2026-07-27
[Feature]: Add integration with GPU-CR


Edwinhr716:
### Problem Statement

Cuda-Checkpoint is slow to do the checkpoint / restore

### Proposed Solution

GPU-CR https://github.com/gpu-os/GPU-CR/tree/main provides faster checkpointing / restore mechanisms. Integration will be split into the following PRs: 

1. API Schema and Python Client update
2. Backend Implementation & Unit tests
3. Helm Chart Update
4. Integration Tests and Documentation

### Willingness to Contribute

Yes, I can submit a PR

PR #141, opened 2026-07-31
[GPU-CR Snapshot-Agent 2/N]: Implementing the backend & unit tests

Edwinhr716:
## What does this PR do?
- Add BackendDirectMemory constant to BackendType
- Implement DirectMemory backend struct using cr_client binary
- Support direct_memory backend in server logic and PID discovery for k8s and standalone modes
- Add BuildDirectMemoryConfig and ExtractDirectMemoryPIDStrings helper functions
- Register DirectMemory backend in cmd/snapshot-agent
- Add unit tests for DirectMemory backend and server integration

## Why is this change needed?
Part of #134 

## How was this tested?

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added direct-memory snapshot and restore support for configured processes.
  * Added direct-memory backend selection for standalone and Kubernetes workflows.
  * Added health checks, PID-based configuration, and configurable operation timeouts.
  * Snapshot and restore operations now support both CUDA and direct-memory configurations.

* **Bug Fixes**
  * Improved backend-specific PID discovery and request routing.
  * Added clearer handling for unavailable tools, invalid process IDs, and command failures.


pkg/snapshot-agent/backends/direct_memory.go
```diff
@@ -0,0 +1,159 @@
...
+	return nil
+}
+
+func (d *DirectMemory) restorePID(ctx context.Context, pid string) error {
+	binaryPath := d.getCrClientPath()
+	if err := d.runCommand(ctx, binaryPath, "-r", "-p", pid); err != nil {
+		return err
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
Edwinhr716: Needs further investigation, created #142

pkg/snapshot-agent/backends/direct_memory.go
```diff
@@ -0,0 +1,165 @@
...
+	} {
+		if path, err := d.lookPath(p); err == nil {
+			return path
+		}
+		if _, err := d.statFunc(p); err == nil {
+			return p
+		}
+	}
+	return "/opt/bin/cr_client"
+}
```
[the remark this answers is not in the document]
Edwinhr716: We control what files are executable, it is fine to not check

pkg/snapshot-agent/backends/direct_memory.go
```diff
@@ -0,0 +1,188 @@
...
+		if err := d.restorePID(ctx, pid); err != nil {
+			return fmt.Errorf("cr_client restore failed for PID %s: %w", pid, err)
+		}
+	}
+	slog.InfoContext(ctx, "cr_client restore took", "duration", time.Since(t0), "pids", pids)
+	return nil
+}
+
+func (d *DirectMemory) getCrClientPath() string {
+	for _, p := range []string{
```
aishukamal: We have full control of the build and installation of the client. So we should only check the exact path that we expect it to be installed in - e.g. `/usr/local/bin/cr_client` and fail hard if it does not exist rather than returning a dangling path.

Also, I assume the cr_client build and install into the snapshot-agent docker image is coming in a separate PR?

pkg/snapshot-agent/backends/direct_memory.go
```diff
@@ -0,0 +1,188 @@
...
+	if out, err := d.execCommand(ctx, name, args...); err != nil {
+		return fmt.Errorf("command failed: %w, output: %s", err, string(out))
+	}
+	return nil
+}
+
+// opTimeout is the per-cr_client-invocation deadline, configurable via
+// GPU_CR_OP_TIMEOUT_SEC (default 120).
+func opTimeout() time.Duration {
+	if v := os.Getenv("GPU_CR_OP_TIMEOUT_SEC"); v != "" {
```
aishukamal: Can we use DIRECT_MEMORY_OP_TIMEOUT_SEC instead? I don't see the env var being defined in main.go.

pkg/snapshot-agent/backends/direct_memory.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+	}
+	return 120 * time.Second
+}
+
+// Both ops pass -b: displacement leaves the process alive and unfrozen —
+// no cuda-checkpoint toggle. Without it, a working toggle freezes CUDA at
+// park and the restore signal handler deadlocks on its first CUDA call
+// while cr_client waits for FINISH before untoggling. (E13 validated this
+// backend as the alternative to cuda-checkpoint, never composed with it.)
```
aishukamal: Looks like crumbs left over from testing - "E13 validated this...". Please remove.

pkg/snapshot-agent/backends/direct_memory.go
```diff
@@ -0,0 +1,188 @@
...
+func (d *DirectMemory) checkpointPID(ctx context.Context, pid string) error {
+	binaryPath := d.getCrClientPath()
+	if err := d.runCommand(ctx, binaryPath, "-c", "-b", "-p", pid); err != nil {
```
aishukamal: I thought `-b` avoids calling cuda-checkpoint after GPU-CR's unmap+release? Curious why we decided not to call cuda-checkpoint and what is the implication of that?

Edwinhr716: yes it does. Claude pushed that change automatically after a test failed. Reverted it to remove the flag, but there is an edge case where cuda-checkpoint fails on restore operations. Haven't found a real life scenario that would run into the edge case so I'm fine with removing `-b`, and I'm investigating further to see if it is worth it to add a fix for the edge case.

pkg/snapshot-agent/backends/direct_memory.go
Edwinhr716: removed

pkg/snapshot-agent/backends/direct_memory.go
Edwinhr716: Changed

pkg/snapshot-agent/backends/direct_memory.go
Edwinhr716: Yes, cr_client build and install comes on a different PR. For now, this backend will remain as not functional.

pkg/snapshot-agent/backends/direct_memory.go
```diff
@@ -0,0 +1,170 @@
...
+	if out, err := d.execCommand(ctx, name, args...); err != nil {
+		return fmt.Errorf("command failed: %w, output: %s", err, string(out))
+	}
+	return nil
+}
+
+// directMemoryOpTimeout is the per-cr_client-invocation deadline,
+// configurable via DIRECT_MEMORY_OP_TIMEOUT_SEC (default 120).
+func directMemoryOpTimeout() time.Duration {
+	if v := os.Getenv("DIRECT_MEMORY_OP_TIMEOUT_SEC"); v != "" {
```
aishukamal, line 103: Please remember to define this env variable in main.go at some point. Not a blocker for this PR.

PR #148, opened 2026-08-12
fix(snapshot-agent): filter out 0-VRAM daemon processes during PID di…

jessicaochen:
## Discovery
Ray worker calling cudaGetDeviceCount to check device count actually spawns a process using 0 GPU memory that the agent detects. Actually checkpointing this process fails due to lack of actual CUDA context.

## Problem

NVML's `GetComputeRunningProcesses()` / `GetGraphicsRunningProcesses()` list non-computing parent/daemon processes (e.g. the Ray worker starter daemon) that merely opened `/dev/nvidiactl` or queried device counts during runtime init. These processes report `UsedGpuMemory == 0` (or `NVML_VALUE_NOT_AVAILABLE`). When snapshot-agent discovers these PIDs and passes them to `cuda-checkpoint --action lock --pid <PID>`, `cuCheckpointProcessLock` fails with `CUDA_ERROR_INITIALIZATION_ERROR` because the process has no active CUDA context to lock.

## Changes

- `getPodPIDsInternal()` in `pkg/snapshot-agent/utils/pod-utils.go` now skips processes whose `UsedGpuMemory` is 0 or `NVML_VALUE_NOT_AVAILABLE`, in both the compute and graphics process loops, before the cgroup membership check.
- Implementation note: the go-nvml constant is `nvml.VALUE_NOT_AVAILABLE` (untyped `-1`) and `UsedGpuMemory` is `uint64`, so the check compares via an `int64` cast: `proc.UsedGpuMemory == 0 || int64(proc.UsedGpuMemory) == nvml.VALUE_NOT_AVAILABLE`.
- Tests: existing mock processes now set a non-zero `UsedGpuMemory` so they still exercise the cgroup-matching path, and a new `Ignore Zero VRAM Processes` case verifies that 0-VRAM and not-available (`^uint64(0)`) processes are filtered out.

## Verification

GOPROXY=off go test -v ./pkg/snapshot-agent/utils/...

All tests pass, including the new `Ignore_Zero_VRAM_Processes` case.

@coderabbitai ignore

### DIFF
--- a/deploy/snapshot-agent/values.yaml
+++ b/deploy/snapshot-agent/values.yaml
@@ -25,6 +25,9 @@ tolerations:
     operator: "Equal"
     value: "present"
     effect: "NoSchedule"
+  - key: "timeslice.io/shared"
+    operator: "Exists"
+    effect: "NoSchedule"
 
 affinity: {}
 
--- a/pkg/snapshot-agent/utils/pod-utils.go
+++ b/pkg/snapshot-agent/utils/pod-utils.go
@@ -41,6 +41,10 @@ var (
 	IsPIDInPodCgroupFunc = isPIDInPodCgroup
 )
 
+// nvmlValueNotAvailable is nvml.VALUE_NOT_AVAILABLE (-1) as it appears in
+// unsigned fields such as ProcessInfo.UsedGpuMemory.
+const nvmlValueNotAvailable = ^uint64(0)
+
 type DeviceInterface interface {
 	GetComputeRunningProcesses() ([]nvml.ProcessInfo, nvml.Return)
 	GetGraphicsRunningProcesses() ([]nvml.ProcessInfo, nvml.Return)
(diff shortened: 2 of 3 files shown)


jessicaochen:
@Edwinhr716 @aishukamal - I agree with claude analysis that this failure has nothing to do with the change. Do let me know if you think otherwise.  (https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/151)

-----------

- CUDACheckpointRestore, VLLMSleepWake, VLLMCompound, VLLMSuspendDiscard, SGLangReleaseResume: PASS
- SGLangCompound: FAIL, same check, in similar time (5.2s)

The strongest signal is that the divergence is byte-identical across both runs:
- before restore: " Paris. It is the seat of the government of France. It is the"
- after restore: " Paris. It is the largest city in Europe and the second largest in the"

Random flakiness (e.g. batching nondeterminism) would produce varying outputs. Instead, the post-restore model deterministically lands in a consistent-but-different state every time. That points at a systematic state corruption or loss in the compound path — SGLang release followed by cuda-checkpoint lock/checkpoint, then restore and resume. Something in what SGLang releases (weights or KV pages) isn't faithfully reconstructed when cuda-checkpoint restore happens on top, while SGLangReleaseResume alone (no cuda-checkpoint layered in) and both VLLM compound variants are fine.

Worth noting: this failure is in code paths untouched by the PR branch (the 0-VRAM filter only changes PID discovery, and both failing runs discovered/locked PIDs successfully) — so this looks like a pre-existing bug in the SGLang compound backend, not a regression from your change.

Edwinhr716 (approved):

PR #154, opened 2026-08-17
snapshot-agent: fix deadlock between StateManager locks

Iceber:
## Why is this change needed?
**Fixing the AB-BA Deadlock in StateManager**

StartSnapshot / StartRestore acquire locks in the order job.mu → sm.mu, while GetJobStatus / UpdateJobPIDs / GetJobPIDs acquire them in the reverse order sm.mu → job.mu — a classic AB-BA deadlock.

The Watcher polls GetJobStatus every second. When that overlaps with a concurrent Snapshot/Restore RPC on the same job, the two goroutines end up waiting on each other indefinitely. Because sm.mu is never released, all subsequent gRPC calls on the agent become blocked, and the only way to recover is to restart the process.

**Fix**
Unify the lock ordering to sm.mu → job.mu. Allow job.mu to be acquired without holding sm.mu, but do not allow sm.mu.Lock to be called while job.mu is held.

## How was this tested?

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved snapshot and restore operation handling to ensure locks are released safely.
  * Prevented potential lock-ordering issues when updating or retrieving job process information.

* **Documentation**
  * Added documentation clarifying lock usage and ordering.

### DIFF
--- a/pkg/snapshot-agent/state-machine/state-manager.go
+++ b/pkg/snapshot-agent/state-machine/state-manager.go
@@ -44,7 +44,10 @@ type Operation struct {
 type StateManager struct {
 	jobs       map[string]*Job
 	operations map[string]*Operation
-	mu         sync.RWMutex
+
+	// mu guards jobs and operations.
+	// Lock order: mu → Job.mu. The reverse order deadlocks.
+	mu sync.RWMutex
 }
 
 // NewStateManager creates a new StateManager instance.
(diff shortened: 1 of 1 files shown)


Iceber:
Only the mutex acquisition order was changed. No backend/GPU/hardware paths are touched. Behavior is identical except for the deadlock fix, and already have unit test coverage."

Iceber:
hi @aishukamal PTAL, Thanks

aishukamal:
Thanks for the fix, @Iceber!

No integration test needed. Label removed.

/approve

PR #162, opened 2026-08-19
client/python: regenerate stubs at floor toolchain, lower dependency floors (v0.2.1)

aishukamal:
The Python client's generated stubs were built with mixed toolchains: orchestrator/health stubs with grpcio-tools 1.81.x (protobuf gencode 6.33.5), snapshot-agent stubs with grpcio-tools 1.83.0 (protobuf gencode 7.35.1). Generated code refuses to import on runtimes older than its gencode, so the snapshot-agent stubs crash on protobuf 6.33 / grpcio 1.81.1 environments, forcing consumers to hand-patch version gates in vendored copies.

This surfaced in the OpenRL integration: OpenRL consumes this client as a git dependency, and its upcoming app-channel integration imports it inside vLLM-based trainer/sampler worker images that run protobuf 6.33 / grpcio 1.81.1 — where the current snapshot-agent stubs fail at import time.

This PR regenerates all stubs with grpcio-tools 1.81.0 (protobuf gencode 6.33.5) using the documented protoc invocations — only version constants change, no descriptor or service changes — and lowers the dependency floors to match: grpcio>=1.81.0, protobuf>=6.33.5. The dev codegen toolchain is pinned so regeneration can't silently raise the floors. Version bumped to 0.2.1.


aishukamal:
Integration test results (GKE, 2x H100 nodes, suites per tests/integration/README.md, images built from this branch):

- snapshot-agent standalone: **PASS 8/8**
- snapshot-agent k8s: **PASS 4/4**
- import matrix — py3.10 + grpcio 1.81.0 / protobuf 6.33.5, and py3.12 + latest grpcio/protobuf: **PASS** (both service stubs import and instantiate at the floor and at latest)
- orchestrator: SingleRLJob **PASS**; QueuedRLJobs **FAIL**

The QueuedRLJobs failure is in the orchestrator→agent cuda-checkpoint preemption path and is unrelated to this PR's Python client changes; it will be followed up separately.

PR #168, opened 2026-08-20
[Snapshot-Agent Direct-Memory 3/N]: Sweep stale GPU-CR artifacts from the shared checkpoint dir

Edwinhr716:
## What does this PR do?
  - Adds `pkg/snapshot-agent/utils/gpu-cr_utils.go`: a sweeper (`StartGPUCRSweeper`) that runs at agent startup and every 10 minutes over the shared GPU-CR checkpoint dir.
- The sweeper deletes:
    - dump/staging pairs (`<id>`, `<id>-host`) that no live process maps (checked against /proc/*/maps) and that are older than 5 minutes;
    - per-PID files (`control-<pid>`, `pid_map_<pid>`) whose process no longer exists.
    - It never touches the bare control counter file or unrecognized files.
  - Adds a startup hook in `cmd/snapshot-agent/main.go,` active only when `EXPORT_FILE_PATH` is set: chmods the checkpoint dir to 0777 (so unprivileged workloads can create files in it) and starts the sweeper.
  - Adds unit tests (`gpu-cr_utils_test.go`) covering the filename patterns and the delete/keep decisions.
  - No new dependencies (stdlib only); no behavior change for deployments without `EXPORT_FILE_PATH`.

## Why is this change needed?

  - GPU-CR leaves per-process artifacts in the shared checkpoint dir (dump/staging pairs, control channels, pid maps) and nothing ever deletes them.
  - The leak is memory, not just disk: the dumps live on hugetlbfs, where `MAP_SHARED` reservations stick to the file a leaked dump pair keeps its full extent (~25 GB on the stock build) of hugepages reserved after the owning process is dead.
  - Hugepages are a fixed pre-carved pool per node, so a couple of leaked dumps exhaust it and every subsequent checkpoint fails with an opaque allocation error. The only remedy today is manually rm-ing files on the node.

  ### Why the snapshot agent and not GPU-CR itself:

  - The files are created by the preloader inside the workload process, and the exits that leak are precisely the ones
    where no cleanup code can run: OOM kills, crashes, pod evictions. A process cannot clean up after its own `SIGKILL`.
  - cr_client can't do it either: it's a short-lived CLI invoked per operation, with no daemon, no session state, and no
    notion of "previous runs". It has no safe moment to decide someone else's files are garbage.
  - Safe deletion requires a node-wide liveness proof: the dir is shared by every GPU-CR tenant on the node, so "is this dump still mapped by any live process?" must be answered across all of `/proc`. Only a long-lived, hostPID, privileged component with the dir mounted can do that, which is exactly the snapshot agent, and nothing on the GPU-CR side has that vantage or lifetime.
  - Upstream GPU-CR also treats artifact lifecycle as the embedder's concern (it standardizes the dir via `EXPORT_FILE_PATH` but ships no reaper), so the deployment that owns the mount owns the hygiene.

## How was this tested?

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for the DirectMemory snapshot backend.
  * Added cleanup for stale GPU checkpoint artifacts across configured export paths.
  * GPU checkpoint directories now use sticky-bit permissions for safer shared access.
* **Bug Fixes**
  * Improved artifact tracking using file identity, preserving active and recently created dumps while removing only confirmed stale files.
* **Tests**
  * Expanded coverage for file-backed dumps, staging files, active mappings, exited processes, and incomplete process scans.


cmd/snapshot-agent/main.go
```diff
@@ -85,6 +87,23 @@ func main() {
...
+	// configured, keeping CUDA/app-only deployments untouched.
+	if ctlDir := os.Getenv("EXPORT_FILE_PATH"); ctlDir != "" {
+		// The dir must be writable by the (unprivileged) GPU-CR workloads
+		// that mmap their dump buffers in it.
+		if _, err := os.Stat(ctlDir); err == nil {
+			if err := os.Chmod(ctlDir, 0o777); err != nil {
+				slog.WarnContext(ctx, "Failed to chmod GPU-CR checkpoint dir to 0777", "dir", ctlDir, "error", err)
+			} else {
+				slog.InfoContext(ctx, "Set GPU-CR checkpoint dir permissions to 0777", "dir", ctlDir)
+			}
```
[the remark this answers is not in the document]
Edwinhr716: Addressed in 5b28134 with the sticky-bit variant: the dir is now chmod'd `1777` (the `/tmp` model), which blocks one tenant from deleting or renaming another's artifacts at the directory level.

Going tighter than world-writable isn't workable here: workload UIDs aren't known up front (arbitrary unprivileged pods), and hostPath volumes get no `fsGroup` remapping, so there's no mechanism to put heterogeneous tenants into a shared group at deploy time. The ownership model is now documented in the code comment at the chmod site. Note per-file exposure is governed by the files' own modes, which GPU-CR controls (its control files are intentionally shared); the sweeper itself runs as root and is unaffected.

pkg/snapshot-agent/utils/gpu-cr_utils.go
```diff
@@ -0,0 +1,118 @@
...
+func liveMappedIds() map[string]bool {
+	ids := make(map[string]bool)
+	procs, err := filepath.Glob("/proc/[0-9]*/maps")
+	if err != nil {
+		return ids
+	}
+	for _, mapsPath := range procs {
+		data, err := os.ReadFile(mapsPath)
+		if err != nil {
+			continue // process exited or unreadable; skip
```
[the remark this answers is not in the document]
Edwinhr716: Fixed in 11bf3d9 as suggested: `liveMappedIds` now returns a `complete` flag that is set false when any maps file fails to read for a reason other than the process exiting mid-scan (`ENOENT`/`ESRCH` stay benign; `EACCES`, hidepid, etc. do not). When the scan is incomplete the sweep logs a warning and skips dump-file deletion for that round — per-PID files are still reaped since their liveness check doesn't use the maps scan. Added `TestSweepIncompleteProcScan` covering the unreadable-maps path (symlink loop → `ELOOP`) and asserting dumps are preserved while dead-PID control files are still removed.

pkg/snapshot-agent/utils/gpu-cr_utils.go
```diff
@@ -0,0 +1,137 @@
...
+// file was unreadable for a reason other than its process exiting mid-scan:
+// a partial scan cannot prove a dump is unmapped, so the caller must skip
+// dump deletion for that sweep.
+func liveMappedIds() (map[string]bool, bool) {
+	ids := make(map[string]bool)
+	complete := true
+	procs, err := filepath.Glob(filepath.Join(procfsRoot, "[0-9]*", "maps"))
+	if err != nil {
+		return ids, false
+	}
```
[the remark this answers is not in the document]
Edwinhr716: Fixed in 984825f as suggested: `liveMappedIds` now enumerates the procfs root with `os.ReadDir`, filters numeric PID dirs (`^\d+$`), and reads each `maps` path directly — so a directory Glob would have silently dropped now surfaces as a read error we can classify. `ENOENT`/`ESRCH` stay benign (process exited between enumerate and read); any other failure, including a ReadDir failure on the root itself, marks the scan incomplete and dump deletion is skipped for that sweep. Tests: the existing symlink-loop case exercises the non-benign path through the new code, and the new `TestLiveMappedIdsExitedProcess` covers the benign path (numeric dir with no `maps` file → scan stays complete).

cmd/snapshot-agent/main.go
```diff
@@ -86,6 +88,27 @@ func main() {
 		backends.BackendDirectMemory: backends.NewDirectMemory(),
 	}
 
+	// GPU-CR housekeeping runs only when the shared checkpoint dir is
+	// configured, keeping CUDA/app-only deployments untouched.
+	if ctlDir := os.Getenv("EXPORT_FILE_PATH"); ctlDir != "" {
```
aishukamal, line 93: IIUC GPU-CR supports a hugepage mode and file mode and EXPORT_FILE_PATH is only set for the file mode. We should cover both, no?

pkg/snapshot-agent/utils/gpu-cr_utils.go
```diff
@@ -0,0 +1,146 @@
...
+//	<id>, <id>-host   dump + staging buffers (hugetlbfs: MAP_SHARED
+//	                  reservations stick to the FILE, so a leaked pair pins
+//	                  the reserved hugepages even after the process dies)
+//	control-<pid>     shared-memory control channel
+//	pid_map_<pid>     pid→id map (may be empty on hugetlbfs)
+//
+// Nothing deletes them on process exit, so the agent sweeps them.
+
+var (
+	dumpFileRe = regexp.MustCompile(`^(\d+)(-host)?$`)
```
aishukamal: Interestingly, this only matches the hugepage-mode names (id, id-host). When EXPORT_FILE_PATH is set, GPU-CR names the files ckpt-<id>.data / ckpt-<id>-host.data which never match. Should we extend the pattern to cover both naming schemes?

pkg/snapshot-agent/utils/gpu-cr_utils.go
```diff
@@ -0,0 +1,146 @@
...
+			// Gone between enumerate and read is normal process churn;
+			// anything else (EACCES, hidepid, ...) hides mappings from
+			// the scan.
+			if !os.IsNotExist(err) && !errors.Is(err, syscall.ESRCH) {
+				complete = false
+			}
+			continue
+		}
+		for _, line := range strings.Split(string(data), "\n") {
+			idx := strings.Index(line, "huge-ckpt/")
```
aishukamal: `/mnt/huge-ckpt` is only used in the hugepage mode, no? In the file mode, the path is whatever EXPORT_FILE_PATH points to. Would it be better to match by file inode instead of name to avoid confusion?

pkg/snapshot-agent/utils/gpu-cr_utils_internal_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"strconv"
+	"testing"
+	"time"
+)
+
+// deadPID is a PID that will not exist on any test host (max pid is
+// typically 4194304 on Linux).
+const deadPID = "999999999"
+
+func TestGCFilePatterns(t *testing.T) {
```
aishukamal, line 17: Let's add a ckpt-id.data / ckpt-id-host.data case for the EXPORT_FILE_PATH mode.

Edwinhr716: Added

cmd/snapshot-agent/main.go
Edwinhr716: Setting`EXPORT_FILE_PATH` doesn't mean that hugepage isn't used. GPU-CR doesn't do any assumptions on whether or not the path is hugepages. So if the directory in `EXPORT_FILE_PATH `is backed by hugepages, then it will use it. 

On some of the features that I added to GPU-CR we do differentiate between hugepages vs not hugepages, but we use `statsfs` and look for `MAP_HUGETLB`.

The plan is to have EXPORT_FILE_PATH always be set when deploying the snapshot-agent daemon for backends that use GPU-CR.

pkg/snapshot-agent/utils/gpu-cr_utils.go
Edwinhr716: Yes, extended to cover both

pkg/snapshot-agent/utils/gpu-cr_utils.go
Edwinhr716: Switched to inode

PR #170, opened 2026-08-21
Add timeslice-slime integration package

aishukamal:
Proposal: https://docs.google.com/document/d/1-JzdzHJ77fZCjgcBVfnANz1z8dipmJWA2vtJ4AzIq2w/edit?tab=t.0#heading=h.dzy9irmry6l

## Summary

Adds the timeslice-slime pip package (pkg/integrations/slime/) — a pre-built PhaseCallback that maps Slime GPU phase boundaries to orchestrator lock acquire/release.

## What is included

| File | Purpose |
|------|---------|
| timeslice_slime/callback.py | TimesliceCallback — acquires/releases orchestrator locks at phase boundaries |
| timeslice_slime/locks.py | RoleLocks — per-role lock primitives with trainer-first ordering, atexit safety, env-gated no-op |
| timeslice_slime/__init__.py | Package root |
| pyproject.toml | Package metadata |
| tests/test_callback.py | 18 lock protocol tests (sync, async, no-op, primitives) |

## How it works

The Slime fork (aishukamal/slime@feat/phase-callbacks) adds --phase-callback-path and emits on_phase_begin/on_phase_end around GPU phases. Users pass --phase-callback-path timeslice_slime.callback.TimesliceCallback — no other Slime code changes needed.

## Test plan

- [x] 18 lock protocol tests pass (python3 -m unittest)
- [x] E2E validated: sync (both pools checkpointed/restored) and async (trainer-only) on H100 cluster

## Summary by CodeRabbit

* **New Features**
  * Added Timeslice integration to coordinate trainer and sampler phases.
  * Added configurable role-based locking with enforced acquisition order, idempotent operations, and safe cleanup.
  * Added no-op behavior when lock configuration is unavailable.
  * Exposed the Timeslice callback for integration use.

* **Tests**
  * Added comprehensive coverage for synchronous and asynchronous workflows, overlapping roles, lock validation, release behavior, and cleanup.

### DIFF
--- a/pkg/integrations/slime/tests/test_callback.py
+++ b/pkg/integrations/slime/tests/test_callback.py
@@ -1,0 +1,364 @@
+"""Lock protocol tests for TimesliceCallback (no GPU, no Slime, no Ray).
+
+Simulates the exact phase sequences that Slime's sync and async drivers
+produce and asserts:
+  * Trainer-first global order (TRAINER never requested while SAMPLER held)
+  * No lock leaks (every acquire has a matching release)
+  * Dual-lock during weight_sync and init (both held during NCCL broadcast)
+  * No-op mode runs clean when env vars are absent
+  * All GPU-touching phases (generate, train, weight_sync, offload, onload,
+    eval, create) are covered by locks
+"""
+
+import importlib.util
+import os
+import unittest
+
+_LOCKS_PATH = os.path.join(os.path.dirname(__file__), "..", "timeslice_slime", "locks.py")
+_spec = importlib.util.spec_from_file_location("_timeslice_locks", _LOCKS_PATH)
+_locks = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(_locks)
+
+SAMPLER = _locks.SAMPLER
+TRAINER = _locks.TRAINER
+RoleLocks = _locks.RoleLocks
+
+JOB = "job-a"
+ADDR = "orch:50051"
+TG = "trainers"
+SG = "samplers"
+
+
+class FakeClient:
+    def __init__(self, events, job_id, group_id):
+        self.events = events
+        self.job_id = job_id
+        self.group_id = group_id
+        self.closed = False
+
+    def acquire(self, group_id, timeout_sec=None):
+        assert group_id == self.group_id
+        self.events.append(("acquire", group_id))
+        return type("R", (), {"waited_ms": 0, "context_restored": False})()
+
+    def release(self, group_id):
+        assert group_id == self.group_id
+        self.events.append(("release", group_id))
+        return type("R", (), {"pending_waiters": 0, "snapshot_deferred": False})()
+
+    def close(self):
+        self.closed = True
+
+
+def make_locks(events):
+    return RoleLocks(
+        job_id=JOB,
+        orch_addr=ADDR,
+        trainer_group=TG,
+        sampler_group=SG,
+        client_factory=lambda target, job_id, group_id: FakeClient(events, job_id, group_id),
+    )
+
+
+def held_after(events):
+    held = set()
+    for op, group in events:
+        if op == "acquire":
+            assert group not in held, f"double acquire of {group}: {events}"
+            held.add(group)
+        else:
+            assert group in held, f"release of un-held {group}: {events}"
+            held.discard(group)
+    return held
+
+
+class CallbackHarness:
+    """Simulate TimesliceCallback without importing Slime."""
+
+    def __init__(self, locks):
+        self.locks = locks
+
+    def on_phase_begin(self, phase, role, context=None):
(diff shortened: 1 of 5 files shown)


pkg/integrations/slime/timeslice_slime/locks.py
```diff
@@ -0,0 +1,174 @@
...
+            raise ValueError(f"unknown role {role!r}")
+        if role == TRAINER and SAMPLER in self._held and TRAINER not in self._held:
+            raise RuntimeError(
+                "lock-order violation: acquiring TRAINER while holding SAMPLER "
+                "(global order is trainer-first; release SAMPLER first)"
+            )
+        if not self.enabled or role in self._held:
+            return
+        t0 = time.monotonic()
+        result = self._clients[role].acquire(group_id=self.groups[role])
```
Edwinhr716: Doesn't this block indefinitely if the lock is never released due to failure? Why not add a timeout_sec parameter to check state periodically, and fail after it has taken too long?

pkg/integrations/slime/timeslice_slime/callback.py
```diff
@@ -0,0 +1,64 @@
...
+            self.locks.acquire(SAMPLER)
+
+    def on_phase_end(self, phase, role, context=None):
+        if role in ("sampler", "both"):
+            self.locks.release(SAMPLER)
+        if role in ("trainer", "both"):
+            self.locks.release(TRAINER)
+
+    def close(self):
+        self.locks.close()
```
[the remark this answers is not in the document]
aishukamal: Overkill for now - the atexit handler is the safety net.

pkg/integrations/slime/timeslice_slime/locks.py
```diff
@@ -0,0 +1,174 @@
...
+        waited_ms = getattr(result, "waited_ms", None)
+        if waited_ms is None:
+            waited_ms = int((time.monotonic() - t0) * 1000)
+        self._held.add(role)
+        _log(
+            f"job={self.job_id} ACQUIRE role={role} group={self.groups[role]} "
+            f"waited={waited_ms}ms context_restored={getattr(result, 'context_restored', '?')}"
+        )
```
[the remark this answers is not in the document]
aishukamal: The orchestrator controls this, not the client.

pkg/integrations/slime/timeslice_slime/locks.py
```diff
@@ -0,0 +1,174 @@
...
+            result = self._clients[role].release(group_id=self.groups[role])
+            _log(
+                f"job={self.job_id} RELEASE role={role} group={self.groups[role]} "
+                f"pending_waiters={getattr(result, 'pending_waiters', '?')} "
+                f"snapshot_deferred={getattr(result, 'snapshot_deferred', '?')}"
+            )
+        except Exception as e:  # noqa: BLE001 - never let a release error kill the job
+            _log(f"job={self.job_id} RELEASE role={role} FAILED: {e}")
+        finally:
+            self._held.discard(role)
```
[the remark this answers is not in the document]
aishukamal: We intentionally keep dependencies = [] since setup_node.sh handles installation order.

pkg/integrations/slime/timeslice_slime/locks.py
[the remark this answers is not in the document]
aishukamal: I was wrong. Went back and looked at the code and the orchestrator does not handle it. We do need to add a timeout and I just did.

pkg/integrations/slime/timeslice_slime/locks.py
aishukamal: Good point. Added a timeout for now. Ideally, this should be fixed server side on the orchestrator, which we can take up separately.

pkg/integrations/slime/tests/test_callback.py
```diff
@@ -0,0 +1,364 @@
...
+            assert group not in held, f"double acquire of {group}: {events}"
+            held.add(group)
+        else:
+            assert group in held, f"release of un-held {group}: {events}"
+            held.discard(group)
+    return held
+
+
+class CallbackHarness:
+    """Simulate TimesliceCallback without importing Slime."""
```
lynnl0927, line 76: If we don't want to import Slime, can we stub slime.utils.phase_callback instead? So we can actually test the real TimesliceCallback and not re-implement it in the tests?

lynnl0927 (approved):

pkg/integrations/slime/tests/test_callback.py
lynnl0927: not blocking

aishukamal: Makes sense to me. Will add a separate PR for this, don't want to hold this PR any further.

PR #176, opened 2026-08-25
Remove node group labels

lynnl0927:
## What does this PR do?

Time-slice groups no longer require operators to label nodes with group.timeslice.io/<group>=true. The orchestrator now derives each group's node membership from where the group's member pods (labeled timeslice.io/group=<group>) are actually scheduled, and groups come into existence when a job first calls acquire(). Physical placement is owned entirely by the Kubernetes scheduler through the shared DRA ResourceClaims the workloads already use. Labeling a node has no effect; existing labels can be removed.

## Why is this change needed?

* Eliminates Manual Toil: Removes operator-applied group.timeslice.io/<group> node labels and fragile nodeSelector boilerplate.
* Pod-Driven Grouping: Workloads declare group membership directly on their pods via timeslice.io/group=<group>. Group node membership is dynamically derived from where live pods land.
* DRA-Driven Placement: DRA claims handle hardware allocation and pod co-location naturally via the Kubernetes scheduler.
* Resilient to Autoscaling & Auto-Repair: GKE node auto-repairs and pool scaling no longer break group discovery due to missing hand-applied node labels.

Followups: 
* Migrate rlts inspect/test-e2e from label scanning to the orchestrator API.
* Automate DRA claim creation.

## How was this tested?

- [x] Integration/e2e tests added/updated

  Official manual integration test results (tests/integration/run.sh --phase orchestrator):

  === RUN   TestOrchestrator
  --- PASS: TestOrchestrator (504.04s)
      --- PASS: TestOrchestrator/SingleRLJob (100.23s)
      --- PASS: TestOrchestrator/QueuedRLJobs (403.39s)
  PASS
  ok  github.com/llm-d-incubation/llm-d-rl-time-slicing/tests/integration/orchestrator  504.063s

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Groups are created automatically on first acquisition.
  * Accelerator placement uses resource claims, with membership discovered from scheduled pods.
  * Workloads no longer require trainer- or sampler-specific node labels and selectors.

* **Bug Fixes**
  * Improved handling of zero-node and temporarily empty groups during startup and recovery.
  * Prevented active groups with lock activity from being removed prematurely.
  * Excluded completed workloads from placement and job tracking.

* **Documentation**
  * Simplified Slime and time-slicing setup, deployment, monitoring, and cleanup instructions.

### DIFF
--- a/guides/rl-frameworks/slime/README.md
+++ b/guides/rl-frameworks/slime/README.md
@@ -40,22 +40,18 @@ Before deploying cooperative time-slicing for Slime, ensure your Kubernetes clus
 * GPU memory capacity must be sufficient to hold the active working set of a single Slime job's trainer or sampler at any one time (since inactive jobs will have their GPU memory checkpointed and evicted).
 * Sampler/trainer node host memory capacity must be sufficient to hold the GPU memory footprint of the trainers/samplers needed for the number of parallel Slime jobs. Worker pod memory limits must account for cuda-checkpoint saving GPU state to host memory — set the pod memory limit to at least 2× the GPU memory footprint of the workload (e.g., 128Gi for a sampler using ~33GB GPU memory with two time-sliced jobs).
 
-### Node Labeling and Tainting for Time-Slice Pools
-The `timeslice` platform relies on node labels and taints to identify resource pools (groups) and isolate time-sliced workloads. For disaggregated Slime executions, label and taint your GPU nodes accordingly:
+### Node Tainting for Time-Slice Pools
+The `timeslice` platform uses a node taint to keep non-time-sliced workloads off shared accelerator nodes:
 
-* **Enable Time-Slicing & Taint Nodes**:
+* **Taint the time-slicing nodes** (prefer node-pool-level taints — hand-applied node metadata does not survive node recreation by auto-repair/upgrade):
   ```bash
-  kubectl label nodes <node-name> timeslice.io/enabled=true
   kubectl taint nodes <node-name> timeslice.io/shared=true:NoSchedule
   ```
-* **Trainer Nodes**:
-  ```bash
-  kubectl label nodes <trainer-node> group.timeslice.io/trainers=true
-  ```
-* **Sampler Nodes**:
-  ```bash
-  kubectl label nodes <sampler-node> group.timeslice.io/samplers=true
-  ```
+
+**No group node labels are needed.** Groups (`samplers`, `trainers`) are created automatically when a job first acquires their lock, and each group's node membership is discovered from where its pods are scheduled. Pod placement itself is driven by the shared DRA `ResourceClaim`s: all jobs' sampler pods reference the samplers claim and therefore co-locate on the claim's allocated GPU (and likewise for trainers), while the scheduler keeps different claims on different devices.
+
+> [!NOTE]
+> `group.timeslice.io/<group>` node labels from older setups are ignored by the platform and can be removed. If your pod templates still carry the matching `nodeSelector`s, ensure the labels exist on nodes or drop the selectors — otherwise the pods will be unschedulable.
 
 ### Shared DRA Resource Claims
 Cooperative time-slicing leverages Kubernetes **Dynamic Resource Allocation (DRA)** so multiple jobs' worker pods can share physical GPU hardware without scheduler blocking. Before submitting jobs, create shared `ResourceClaim` manifests for both the trainer and sampler pools in your target namespace:
@@ -214,7 +210,7 @@ When deploying Slime across independent Ray clusters, use KubeRay `RayJob` manif
 
 A complete disaggregated Slime workload requires defining **two separate worker groups** under `workerGroupSpecs`: one for trainers and one for rollouts (samplers). For each group:
 * **Custom Ray Resources**: Include custom resource counts in `rayStartParams` (`"{\"trainers\": 1}"` for trainers and `"{\"samplers\": 1}"` for rollouts) so that Ray placement groups can bind tasks to the appropriate worker pool.
-* **Pool-Specific Identifiers**: Ensure each worker group is configured with its corresponding node selector (`group.timeslice.io/trainers: "true"` vs. `samplers: "true"`), pod labels (`timeslice.io/group: trainers` vs. `samplers`), and shared DRA claim (`shared-trainers-gpu-claim` vs. `shared-samplers-gpu-claim`).
+* **Pool-Specific Identifiers**: Ensure each worker group is configured with its corresponding pod labels (`timeslice.io/group: trainers` vs. `samplers`) and shared DRA claim (`shared-trainers-gpu-claim` vs. `shared-samplers-gpu-claim`). The claim determines placement — no node selector is required.
 * **Time-Slicing Environment Variables**: Set `TIMESLICE_JOB_ID`, `TIMESLICE_ORCH_ADDR`, `TIMESLICE_TRAINER_GROUP`, `TIMESLICE_SAMPLER_GROUP`, and NCCL env vars in `runtimeEnvYAML`.
 
 For **async RL**, set `TIMESLICE_SAMPLER_GROUP` to a per-job value (e.g., `samplers-${JOB_NAME}`) so each job gets its own sampler group with no contention.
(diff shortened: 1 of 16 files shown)


guides/rl-frameworks/slime/examples/async/README.md
```diff
@@ -17,32 +17,21 @@ In async mode, generation N+1 runs concurrently with training N. Sampler GPUs ar
 
 All files (launch script, RayJob template, setup script, resource claims) are shared with the sync example — only `SLIME_MODE` differs.
 
-**Note:** For async mode, the sampler node needs per-job group labels so the orchestrator can discover each job's sampler group:
-```bash
-kubectl label nodes <sampler-node> group.timeslice.io/samplers-slime-async-1=true group.timeslice.io/samplers-slime-async-2=true
-```
+**Note:** No node labels are needed for the per-job sampler groups — each group (`samplers-${JOB_NAME}`) is created automatically when the job first acquires its lock, and its node membership is discovered from where the job's sampler pods are scheduled (placement follows the shared sampler `ResourceClaim`).
```
aishukamal, line 20: If this is completely transparent to the end user, do we even need to mention anything about node group labels?

guides/rl-frameworks/slime/README.md
```diff
@@ -40,22 +40,18 @@ Before deploying cooperative time-slicing for Slime, ensure your Kubernetes clus
...
-  ```
-* **Sampler Nodes**:
-  ```bash
-  kubectl label nodes <sampler-node> group.timeslice.io/samplers=true
-  ```
+
+**No group node labels are needed.** Groups (`samplers`, `trainers`) are created automatically when a job first acquires their lock, and each group's node membership is discovered from where its pods are scheduled. Pod placement itself is driven by the shared DRA `ResourceClaim`s: all jobs' sampler pods reference the samplers claim and therefore co-locate on the claim's allocated GPU (and likewise for trainers), while the scheduler keeps different claims on different devices.
+
+> [!NOTE]
+> `group.timeslice.io/<group>` node labels from older setups are ignored by the platform and can be removed. If your pod templates still carry the matching `nodeSelector`s, ensure the labels exist on nodes or drop the selectors — otherwise the pods will be unschedulable.
```
aishukamal, line 54: Same here. Given that at this stage in the project we don't need to worry about backwards compatibility (no real customers), I wonder if we can remove any mention of node group labels altogether from all guides. The guides should look as if this was the design from the get go.

guides/rl-frameworks/slime/README.md
```diff
@@ -40,22 +40,18 @@ Before deploying cooperative time-slicing for Slime, ensure your Kubernetes clus
...
 
-### Node Labeling and Tainting for Time-Slice Pools
-The `timeslice` platform relies on node labels and taints to identify resource pools (groups) and isolate time-sliced workloads. For disaggregated Slime executions, label and taint your GPU nodes accordingly:
+### Node Tainting for Time-Slice Pools
+The `timeslice` platform uses a node taint to keep non-time-sliced workloads off shared accelerator nodes:
 
-* **Enable Time-Slicing & Taint Nodes**:
+* **Taint the time-slicing nodes** (prefer node-pool-level taints — hand-applied node metadata does not survive node recreation by auto-repair/upgrade):
   ```bash
-  kubectl label nodes <node-name> timeslice.io/enabled=true
```
aishukamal, line 48: We still need this, don't we?

guides/timeslice-orchestrator/README.md
```diff
@@ -72,9 +72,9 @@ To enable time-slicing, the cluster must be configured with:
 - **Kubernetes Version:** 1.34 or later (required for DRA v1).
 - **At least one GPU node with:**
   - **NVIDIA Driver:** 565 or later (required for DRA Driver).
-  - **Labels & Taints:** `timeslice.io/enabled=true` and tainted with `timeslice.io/shared=true:NoSchedule` to isolate time-slicing workloads.
```
aishukamal, line 75: We still need this?

guides/rl-frameworks/slime/examples/ray-job.yaml.template
```diff
@@ -86,8 +86,6 @@ spec:
             timeslice.io/job-id: "${JOB_NAME}"
             timeslice.io/group: "trainers"
         spec:
-          nodeSelector:
-            group.timeslice.io/trainers: "true"
```
aishukamal, line 90: I think the nodeSelector should be `timeslice.io/enabled: "true"`. This guide had it wrong to begin with. Without this, we would not be able to run time-sliced and non-time-sliced workloads on the same cluster since snapshot-agent would only run on nodes with this label.

guides/rl-frameworks/slime/examples/ray-job.yaml.template
```diff
@@ -145,8 +143,6 @@ spec:
             timeslice.io/job-id: "${JOB_NAME}"
             timeslice.io/group: "${SAMPLER_TS_GROUP}"
         spec:
-          nodeSelector:
-            group.timeslice.io/samplers: "true"
```
aishukamal, line 149: Same here. Replace with `timeslice.io/enabled: "true"`.

guides/rl-frameworks/slime/examples/async/README.md
aishukamal, line 1: Would there be similar updates needed to the sync README and examples? I'd be fine with them coming in a follow-up PR but let's make sure they are not missed.

Separately, we should vet every guide/example/code and make sure they don't have a dependency on the stale group labels. For example, I believe the `rlts inspect` command relies on group labels, so we'll need to fix those.

tests/integration/orchestrator/harness.go
```diff
@@ -110,9 +110,16 @@ func NewComposedHarness(t *testing.T) *ComposedHarness {
...
+	// Label each group node exclusively — unless the run validates label-free
+	// membership (INTEG_EMERGENT set), in which case NO node labels may
+	// exist: the orchestrator must create groups on first Acquire and
+	// discover their nodes from scheduled pods.
+	if os.Getenv("INTEG_EMERGENT") == "" {
+		h.exclusiveLabel(t, samplerNode, integSamplers)
+		h.exclusiveLabel(t, trainerNode, integTrainers)
+	} else {
+		t.Log("INTEG_EMERGENT set: skipping group node labeling (label-free membership mode)")
+	}
```
aishukamal, line 122: If the user explicitly labels nodes, I don't see how that would work, since the orchestrator doesn't read node labels anymore. Can we make the no-group-label mode the default and only mode? Can be done in a separate PR.

aishukamal (commented):
Overall, this is a welcome change. Most of my comments are about updating the relevant artifacts (guides, runnable examples, tools, integration tests) to stop using node group labels altogether. That's a separate clean-up effort and I'm fine if you'd like to do that in a separate PR.

jessicaochen (changes requested):
Overall, it is good to not fix group membership to the nodes itself. However, to do that properly there are a few high level things that need to work:
* use pod anti-affinity on your group key to provide the group isolation that used to be provided at the node level (needed docs, e2e testing & slime example) 
  * We currently only support nodes belonging to one group at a time. Since we are no longer enforcing that with labels, you will need to update documentation & examples so it is enforced at pod level.
  * DRA itself has no node selectors so it cannot be used to keep your samplers away from your trainers. 
* Pods using GPUs that are not configured for timeslicing should not be sharing a node with pods that are using timeslicing (we presume ones configued for timeslicing will lock and to that properly. ones not configured will probably not grab locks appropriately). This is why that general timeslice.io/enabled: "true" label, taint & toleration existed. If you decide to remove that label, taint & toleration you will need to find an alternate way to keep the non-timeslicing pods away from the timeslicing ones.  (I will note that taint/toleration is the best way I can think of. Cannot really do antiaffinity on key blanks + resources).
* You will want to make sure it works with the slime example before submitting this PR. That is a good test we have not forgotten something since it is a real RL framework.

I'll swing back by for a more detailed review after your response to the above high level stuff.

guides/rl-frameworks/slime/examples/ray-job.yaml.template
jessicaochen: This section is actually provisioning the ray worker pod. Since the nodes are labeled with both  ```group.timeslice.io/trainers: "true"``` & ```timeslice.io/enabled: "true"``` both agent and the worker pod run on the node. You are correct though that if we remove ```group.timeslice.io/trainers: "true"``` then ```timeslice.io/enabled: "true"``` becomes required if we do not want the pods to run off and run on machines without ```timeslice.io/enabled: "true"```.

PR #171, opened 2026-08-21
[Snapshot-Agent Direct Memory 4/N]: Python Library Config Helpers

Edwinhr716:
## What does this PR do?

Adding config builders to the python client

## Why is this change needed?

Makes it easier to build the different backend configs we support

## How was this tested?

- [X] Unit tests added/updated
- [ ] Integration/e2e tests added/updated
- [ ] Manual testing performed

## Checklist

- [ ] Commits are signed off (`git commit -s`) per [DCO](PR_SIGNOFF.md)
- [ ] Code follows project [contributing guidelines](CONTRIBUTING.md)
- [ ] Tests pass locally (`make test`)
- [ ] Linters pass (`make lint`)
- [ ] Documentation updated (if applicable)

PR #187, opened 2026-09-01
[Snapshot-Agent]: Add config builders for AppEndpointConfig and AppChannelConfig

Edwinhr716:
## What does this PR do?
Adds config builders for other backends, matching #171

## Why is this change needed?

See https://github.com/llm-d-incubation/llm-d-rl-time-slicing/pull/171#discussion_r3848725490

## How was this tested?

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added configuration builders for application endpoints and channels.
  * Added ready-to-use configuration presets for vLLM and SGLang applications.
  * Added support for configuring suspend modes and region tags using friendly names or numeric values.
  * Added validation for applications, endpoints, suspend modes, tags, and configuration fields.
  * Expanded the public Python API to expose the new configuration helpers.

### DIFF
--- a/pkg/client/python/timeslice/snapshot_agent/configs.py
+++ b/pkg/client/python/timeslice/snapshot_agent/configs.py
@@ -44,6 +44,8 @@ def _suspend_mode(mode: Union[str, int, None]) -> int:
 
 
 def _tags(tags: Optional[Sequence[str]]) -> list:
+    if isinstance(tags, str):
+        raise ValueError("tags must be a sequence of non-empty strings")
     tags = list(tags or [])
     if any(not isinstance(t, str) or not t for t in tags):
         raise ValueError(f"tags must be non-empty strings, got {tags!r}")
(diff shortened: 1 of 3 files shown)


pkg/client/python/timeslice/snapshot_agent/configs.py
```diff
@@ -16,6 +16,42 @@ def _process_target(pids: Sequence[int]) -> snapshot_agent_pb2.ProcessTarget:
...
+            return _SUSPEND_MODES[mode.lower()]
+        except KeyError:
+            raise ValueError(
+                f"unknown suspend mode {mode!r}; expected one of {sorted(_SUSPEND_MODES)}"
+            )
+    snapshot_agent_pb2.SuspendMode.Name(mode)  # raises ValueError if invalid
+    return mode
+
+
+def _tags(tags: Optional[Sequence[str]]) -> list:
```
jessicaochen, line 46: Interesting that the tags are strings while most of the other settings are enums. Are arbitrary values allowed here or only certain values legit? If the latter, should we do some validation that we are not being passed invalid or unsupported tags?

jessicaochen: Nevermind, looks like tests cover it which means it is probably a different part of the code that is rejecting invalid values.

jessicaochen (approved):

PR #189, opened 2026-09-01
[GPU-CR Selective Checkpoint 4/N]: cr_client support for selective checkpoint/restore of GPU memory regions

Edwinhr716:
## How to call selective checkpoint

  The target process must be running with `vGPU.so` preloaded and initialized
  (`-i` arms the handlers and publishes selective support). The process **keeps
  running** during selective ops — there is no `cuda-checkpoint` freeze/thaw.

  ```sh
  # One-time per process: initialize the preloaded library
  cr_client -i -p <pid>

  # Checkpoint specific device regions into the per-PID staging buffer
  cr_client -c -p <pid> -s 0x7f0000000000:4096,0x7f0000100000:8192

  # Checkpoint the same regions to a file instead
  cr_client -c -p <pid> -s <ptr:size,...> -o /store/dump.bin

  # Restore, from the buffer or from a file
  cr_client -r -p <pid> -s <ptr:size,...> [-o /store/dump.bin]
  ```

  - `-s` takes comma-separated `ptr:size` pairs, hex or decimal, up to 4095
    regions per request. Malformed or overflowing values fail at parse time.
  - `-o` must be an absolute path. After a checkpoint the client validates the
    dump's commit marker; before a restore it refuses a torn or empty dump.
  - Every op has a deadline: `GPU_CR_OP_TIMEOUT_SEC` (default 120 s).

  ## Exit codes

  `cr_client` used to exit 0 or 1. Codes 2–4 are new, so callers (the snapshot
  agent) can tell outcomes apart:

  | Exit | Meaning | Typical cause |
  |------|---------|---------------|
  | 0 | Success | |
  | 1 | Usage error | bad flag combination, malformed `-s`, over-long `-o` |
  | 2 | Op failed | the library reported failure, dump failed validation, bad destination path |
  | 3 | Refused pre-signal | library has not published selective support, or the target died — nothing was perturbed |
  | 4 | Timeout | workload never finished the op; restart the workload before issuing another op on that PID |

## What does this PR do?

  Adds the coordinator-side client for the selective checkpoint protocol that
  landed in #177, plus hardening of the shared `cr_client` path:

  - **Selective ops**: `-s` (region list) and `-o` (file destination) drive
    `SELECTIVE_CKPT_MSG` / `SELECTIVE_RESTORE_MSG`. Region parsing is strict
    (`selective_spec.h`, header-only), fails at admission time, and enforces the
    region-count bound so a request can never fail for table capacity after the
    device-to-host copy already ran.
  - **Safety around the shared channel**: a `flock` on the control file
    serializes concurrent clients for the whole op; selective ops are refused
    before any signal if the library never published support; destination files
    are pre-created with symlink protection and truncation so a recycled path
    can never make a torn dump look valid, then chown'd to the target's uid
    with mode 0600 — the workload (often non-root) writes the dump, and
    owner-only write keeps a co-mounted uid from planting a marker-valid dump
    that a restore would trust.
  - **Hardening of existing full ops**: a dead target fails immediately instead
    of burning the timeout; the previously unbounded finish-wait now has a
    deadline; `cuda-checkpoint` runs via fork/exec instead of `system()` (works
    in images with no `/bin/sh`); a failed freeze or thaw is reported instead of
    silently ignored.
  - **`multi_cr_client`** sizes worker mappings by `fstat` instead of a
    compile-time constant, removing a cross-binary size coupling.

## Why is this change needed?

  The snapshot agent needs to move *chosen* memory regions (one model's weights,
  one group's KV) in and out of a live process. A full checkpoint freezes the
  process and copies all of VRAM; selective checkpoint copies only the named
  regions while the process keeps serving. #177 defined the wire format; this PR
  adds the client that actually drives it — without it the protocol is
  unreachable.

  The hardening exists because a machine, not a human, runs this binary: the
  agent must be able to distinguish "done", "failed", "refused, nothing
  happened", and "wedged" from the exit code alone, and must never be told a
  process parked when it is still running on the GPU. Before this change,
  `cr_client` could hang forever on a wedged workload and silently mis-report
  toggle failures.

## How was this tested?

- [X] Integration/e2e tests added/updated

## Checklist

### DIFF
--- a/third_party/gpu-cr/coordinator/cr_client.cpp
+++ b/third_party/gpu-cr/coordinator/cr_client.cpp
@@ -1,7 +1,9 @@
 #include <iostream>
 #include <stdint.h>
 #include <unistd.h>
+#include <errno.h>
 #include <fcntl.h>
+#include <sys/file.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <assert.h>
@@ -12,15 +14,34 @@
 #include <string>
 
 #ifdef __HIP_PLATFORM_AMD__
-// AMD platform 
+// AMD platform
 #else
 #include <cuda.h>
 #endif
 
 #include "common.h"
+#include "dump_format.h"
+#include "selective_spec.h"
 #include "comm/comm.h"
 
+namespace {
+
+// Exit codes (consumed by the snapshot agent):
+//   0 OK, 1 usage/parse, 2 op failed (op_status or invalid dump),
+//   3 gate refused (capability), 4 timeout.
+constexpr int kExitOpFailed = 2;
+constexpr int kExitRefused = 3;
+constexpr int kExitTimeout = 4;
+
+}  // namespace
+
 std::string get_cuda_checkpoint_path() {
+    // Deployment override (also used by the GPU-free integration tests):
+    // point at the cuda-checkpoint binary explicitly instead of relying on
+    // the layout-relative lookup below.
+    const char* env_path = getenv("GPU_CR_CUDA_CHECKPOINT");
+    if (env_path && env_path[0]) return env_path;
+
     char exe_path[1024];
     ssize_t count = readlink("/proc/self/exe", exe_path, 1024);
     if (count == -1) {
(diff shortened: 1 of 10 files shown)


third_party/gpu-cr/coordinator/cr_client.cpp
```diff
@@ -34,12 +58,165 @@ std::string get_cuda_checkpoint_path() {
...
+    // defended — the agent passes store paths whose parents it owns; a
+    // workload-writable intermediate would need
+    // openat2(RESOLVE_NO_SYMLINKS) to close.
+    int fd = openat(dfd, base,
+                    O_CREAT | O_RDWR | O_TRUNC | O_CLOEXEC | O_NOFOLLOW, 0644);
+    close(dfd);
+    if (fd < 0) {
+        fprintf(stderr, "Error: cannot create destination %s: %s\n", path, strerror(errno));
+        return false;
+    }
```
aishukamal, line 156: Do we need a `fchmod(fd, 0666);` after this block since the file is created  by cr_client running as root on snapshot agent and the workload (likely running non-root) will not have write access to the file?

Edwinhr716: Good catch — this was a real bug, though I went a slightly different way than `fchmod 0666`. The failure you describe is exactly right: the .so opens `dest_path` with `O_RDWR` (never `O_CREAT` — existence is the client's responsibility so a stale path can't materialize files), so a root-owned 0644 precreate turns every checkpoint-to-file into EACCES inside a non-root workload.

Fixed in 9e09574 by handing the inode to the target instead of opening it up: `SecurePrecreate` now `fchown`s to the target's uid/gid (stat of `/proc/<pid>`, so no cross-pod UID contract is needed) and pins mode 0600 with an explicit `fchmod` (umask-independent). The reason I avoided 0666: the store dir is shared, and `ValidateDestDump` proves a commit marker, not authorship — a world-writable dump would let any co-mounted uid overwrite it with marker-valid bytes that a later restore trusts. Owner-rw keeps write access scoped to the process whose memory the dump actually is; the validating reopen runs as root so it's unaffected. A non-dumpable target stats as root:root and just keeps today's behavior.

The integration test now asserts owner == target and mode 600 after a dest-path checkpoint (without the fchmod a default umask leaves 644, so the mode assertion alone pins the handoff), and the case matrix has the row.

Edwinhr716: Added `fchmod(fd, 600)` to give write permissions to the pid that where the checkpoint / restore is being done, to minimize permissions given.

aishukamal (approved):

PR #192, opened 2026-09-02
[Snapshot-Agent Memory-Regions 3/N]: Implementing the backend and unit tests

Edwinhr716:
## What does this PR do?

It adds the `memory_regions` backend to the snapshot-agent: selective checkpoint/restore of caller-listed device-memory regions via `cr_client` destination-path operations, stored in named per-slot directories and garbage-collected by owner liveness. The full design can be found here https://docs.google.com/document/d/1qAGiKgmb_5v8J4B-c6qPp6OzmubLEmZrlURVWMP8h04/edit?tab=t.0 

  **Where things live:**
  - `backends/memory_regions.go` — the backend: validation, slot handling,`ensureIDForPid`, the per-PID `cr_client` invocations, `.owners` writing (with the hugetlbfs EINVAL fallback).
  - `utils/gpu-cr_utils.go` — the sweep side, extending the existing GPU-CR sweeper: the group-store owner-liveness pass, metadata fallback reads, orphan reaping, and a shared `StoreMu` so sweeps and operations never interleave on the store.
  - `server/server.go` — routing: memory-regions requests go through the slot-aware state transitions from #165; the config passes through untouched; region PIDs are cached for `Status` parity.
  - `cmd/snapshot-agent/main.go` — registration. Requests stay behind the`MemoryRegionsBackend` feature gate from #163 (default off), so merging this changes nothing for existing deployments.

  **What's deliberately *not* here:** no Helm/deploy changes, no PID or region discovery, and no Python client helpers.

## Why is this change needed?

The API (#163) and the slot state machine (#165) are already merged, but a memory-regions request currently routes to a backend that doesn't exist. This PR is the missing piece that makes the feature real.

## How was this tested?

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for snapshotting and restoring selected device-memory regions.
  - Added named snapshot slots, with job-based defaults when no name is provided.
  - Added validation for invalid regions, processes, and snapshot names.
  - Added health checks for required checkpoint tooling.
  - Added support for tracking memory-region ownership across snapshot operations.

- **Bug Fixes**
  - Improved cleanup by retaining artifacts while owning processes remain active.
  - Added safeguards for recycled process IDs and incomplete process information.
  - Improved handling of missing snapshots and checkpoint operation failures.


pkg/snapshot-agent/backends/memory_regions.go
```diff
@@ -0,0 +1,437 @@
...
+
+// groupDir validates a snapshot slot name and returns its directory under
+// the destination store. Slots must be a single path segment: nesting is
+// rejected along with traversal because GC reaps store entries at the top
+// level only (a nested slot's .owners metadata would be invisible to it).
+func groupDir(slot string) (string, error) {
+	store := utils.GroupStoreDir()
+	dir := filepath.Join(store, filepath.Clean(slot))
+	rel, err := filepath.Rel(store, dir)
+	if err != nil || rel == "." || strings.HasPrefix(rel, "..") || strings.ContainsRune(rel, os.PathSeparator) {
```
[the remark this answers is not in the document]
Edwinhr716: Intentional strictness rather than a bug: `groupDir` is the single path-security chokepoint for caller-controlled slot names, and over-rejection is the safe direction there. A name beginning with `..` has no legitimate caller (slots are adapter/job names), so permitting `..slot` buys nothing while making the guard's predicate subtler to reason about. Keeping the simpler, stricter check.

pkg/snapshot-agent/backends/memory_regions.go
```diff
@@ -0,0 +1,437 @@
...
+	for _, pid := range regionPIDs(cfg) {
+		pidStr := strconv.Itoa(int(pid))
+		id, err := g.ensureIDForPid(ctx, pidStr)
+		if err != nil {
+			return fmt.Errorf("failed to resolve PID %d to ID: %w", pid, err)
+		}
+		dest := filepath.Join(targetDir, id)
+		specStr := strings.Join(specs[pid], ",")
+		if err := g.checkpointRegions(ctx, pid, specStr, dest); err != nil {
+			return fmt.Errorf("cr_client checkpoint failed for PID %d with specs %s: %w", pid, specStr, err)
```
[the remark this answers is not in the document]
Edwinhr716: Valid — fixed in 060b805, with a lighter mechanism than a staging slot: the owners metadata is now written right after the slot directory is created, before any checkpoint runs. Every slot that exists carries its owners from the start, so a partial multi-PID failure leaves a slot the sweeper reclaims normally once the owners die (and a retry overwrites in place while they live). Staging+publish would add a rename step to paths the workload writes concurrently, for no additional safety. Regression test: TestMemoryRegionsSnapshotPartialFailureKeepsOwners.

pkg/snapshot-agent/utils/gpu-cr_utils.go
```diff
@@ -88,26 +162,228 @@ func sweep(ctlDir string) {
...
+		}
+		owners, err := ReadGroupMeta(dir)
+		if err != nil || len(owners) == 0 {
+			// No metadata: be conservative, never delete.
+			continue
+		}
+		allGone := true
+		for pid, st := range owners {
+			cur, err := ProcStarttime(pid)
+			if err == nil && cur == st {
```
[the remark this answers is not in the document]
Edwinhr716: Valid — fixed in b2daeac. The owner check now mirrors pidGone's conservative contract via a shared-shape helper (ownerGone): on a starttime read error the owner counts as gone only when the process is confirmed absent; unreadable-or-malformed stat with /proc/<pid> still present keeps the slot. Covered by TestOwnerGone.

pkg/snapshot-agent/utils/gpu-cr_utils.go
```diff
@@ -1,6 +1,7 @@
 package utils
```
jessicaochen: You should not have a package called utils. This is a bad name because it does not bound the contents of the package so it tends to grow indefinitely & become unmanageable. Looking at the files you are touching, perhaps a gpu-cr package makes sense?

pkg/snapshot-agent/server/server.go
```diff
@@ -225,6 +238,39 @@ func (s *Server) buildSnapshotFn(
 	}
 }
 
+// memoryRegionsSlot returns the snapshot slot a memory-regions request
+// targets (snapshot_name, defaulting to the job ID), or "" for any other
+// backend config. Slot-aware state transitions (live slot swap, fault
+// recovery) only apply when this is non-empty.
+func memoryRegionsSlot(config *pb.BackendConfig, jobID string) string {
```
jessicaochen, line 245: Where are the unit test for this new logic? I presume these helper functions at minimum are testable.

pkg/snapshot-agent/utils/gpu-cr_utils.go
```diff
@@ -41,10 +50,68 @@ func isDumpFile(name string) bool {
...
+	idx := strings.LastIndexByte(string(data), ')')
+	if idx < 0 || idx+2 >= len(data) {
+		return 0, fmt.Errorf("malformed /proc/%s/stat", pid)
+	}
+	fields := strings.Fields(string(data[idx+2:]))
+	// fields[0] is field 3 (state); starttime is field 22.
+	if len(fields) < 20 {
+		return 0, fmt.Errorf("short /proc/%s/stat", pid)
+	}
+	return strconv.ParseInt(fields[19], 10, 64)
```
jessicaochen: Lots of magic numbers in this function, use local consts for readability.

pkg/snapshot-agent/backends/memory_regions.go
```diff
@@ -0,0 +1,434 @@
...
+// with no hugepages-2Mi request.
+//
+// Environment configuration:
+//
+//	EXPORT_FILE_PATH        GPU-CR data dir: dump/staging buffers and the
+//	                        destination group store (default /mnt/huge-ckpt)
+//	GPU_CR_CTL_PATH         control-plane tmpfs (control-<pid>,
+//	                        pid_map_<pid>, ctl-ready-<pid>); unset = legacy
+//	                        layout sharing the data dir
+//	GPU_CR_GROUP_STORE      destination store override (default <data>/groups)
```
jessicaochen, line 39: Recommend your ENV variable names be consts instead of hardcoded in multiple places. That way if you need a rename etc, it automatically happens in all the places.

pkg/snapshot-agent/backends/memory_regions.go
```diff
@@ -0,0 +1,434 @@
...
+	}
+	specs := make(map[int32][]string)
+	for _, r := range regions {
+		if r.GetPid() <= 0 {
+			return nil, fmt.Errorf("memory region pid must be positive, got %d", r.GetPid())
+		}
+		if r.GetSizeBytes() == 0 {
+			return nil, fmt.Errorf("memory region size_bytes must be positive (pid %d, address 0x%x)", r.GetPid(), r.GetAddress())
+		}
+		specs[r.GetPid()] = append(specs[r.GetPid()], fmt.Sprintf("0x%x:%d", r.GetAddress(), r.GetSizeBytes()))
```
jessicaochen, line 142: If it is possible to list a particular pid multiple times, do we need to ensure there are no duplicate address/size pairs?

pkg/snapshot-agent/utils/gpu-cr_utils.go
```diff
@@ -63,16 +130,27 @@ func StartGPUCRSweeper(ctx context.Context, ctlDir string, interval time.Duratio
...
 
-func sweep(ctlDir string) {
-	entries, err := os.ReadDir(ctlDir)
+func sweep(dataDir string) {
+	sweepDataDir(dataDir)
+	if ctl := os.Getenv("GPU_CR_CTL_PATH"); ctl != "" && ctl != dataDir {
+		sweepPidFiles(ctl, ctlMinAge)
+	}
+
+	StoreMu.Lock()
```
jessicaochen: recommend to use defer to unlock. It will keep the logic safe even if more logic is added to the area under lock

jessicaochen (changes requested):

pkg/snapshot-agent/utils/gpu-cr_utils.go
```diff
@@ -88,26 +166,196 @@ func sweep(ctlDir string) {
...
+		}
+		m := ownerDirRe.FindStringSubmatch(e.Name())
+		if m == nil {
+			return nil, false
+		}
+		st, err := strconv.ParseInt(m[2], 10, 64)
+		if err != nil {
+			return nil, false
+		}
+		owners[m[1]] = st
```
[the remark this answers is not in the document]
Edwinhr716: slotOwners now returns one record per parsed <pid>-<starttime> dirname instead of a pid-keyed map, and the sweep requires every record to be gone. Regression coverage: TestSlotOwners parses the same-pid/two-starttimes pair as distinct records, and the sweep test adds a "recycled-duplicate" slot whose dead dirname sorts after the live one, exactly the ordering that made the old map retain the stale starttime, and asserts it is kept.

pkg/snapshot-agent/utils/gpu-cr_utils.go
```diff
@@ -88,26 +166,196 @@ func sweep(ctlDir string) {
...
+// ownerGone reports whether a recorded slot owner (pid + starttime) is gone:
+// the process exited, or its pid now belongs to a different process
+// (starttime mismatch). Liveness read errors are treated conservatively,
+// mirroring pidGone: while the process may still exist, the owner counts as
+// alive — a destination slot is the sole copy of parked state, so deletion
+// requires positive evidence of death, never the absence of evidence.
+func ownerGone(pid string, recorded int64) bool {
+	cur, err := ProcStarttime(pid)
+	if err != nil {
+		return os.IsNotExist(err) || !procExists(pid)
```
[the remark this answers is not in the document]
Edwinhr716: An owner now counts as gone only on positive evidence: starttime read errors delete only when a follow-up stat confirms /proc/<pid> is absent; permission or I/O errors on either read keep the slot. The doc comment now states this is deliberately stricter than pidGone, whose ctl-file deletions are recoverable.

pkg/snapshot-agent/server/server.go
Edwinhr716: Added unit tests

pkg/snapshot-agent/backends/memory_regions.go
Edwinhr716: Changed to consts

pkg/snapshot-agent/backends/memory_regions.go
Edwinhr716: It is possible to list a particular pid multiple times, but duplicate address does not make sense. Added check and reject duplicate addresses.

pkg/snapshot-agent/utils/gpu-cr_utils.go
Edwinhr716: Moved lock to be inside sweepGroupStore and deferred the unlock there instead

pkg/snapshot-agent/utils/gpu-cr_utils.go
Edwinhr716: added local consts

pkg/snapshot-agent/utils/gpu-cr_utils.go
Edwinhr716: moved to its own package

pkg/snapshot-agent/backends/memory_regions_test.go
```diff
@@ -0,0 +1,635 @@
...
+// inside per-owner <pid>-<starttime> dirs with a pinned fake starttime
+// (test pids have no procfs entry).
+//
+//nolint:gocritic // The project configuration bans named returns, conflicting with unnamedResult
+func newMemoryRegions(t *testing.T) (*backends.MemoryRegions, string, string) {
+	t.Helper()
+	ctlDir := t.TempDir()
+	t.Setenv("EXPORT_FILE_PATH", ctlDir)
+	t.Setenv("GPU_CR_CTL_PATH", "")
+	t.Setenv("GPU_CR_GROUP_STORE", "")
```
jessicaochen: (nit) should this be using the consts

jessicaochen: True for several tests.

pkg/snapshot-agent/backends/memory_regions.go
```diff
@@ -0,0 +1,447 @@
...
+// The agent moves NO dump bytes itself: each snapshot is
+// written by the workload's preloader directly into a per-slot destination
+// file (cr_client -o), and each restore reads straight from it. The agent
+// only names files, pre-creates them (via cr_client), and garbage-collects —
+// none of which faults a hugetlb page, which is what lets the DaemonSet run
+// with no hugepages-2Mi request.
+//
+// Environment configuration:
+//
+//	EXPORT_FILE_PATH        GPU-CR data dir: dump/staging buffers and the
```
jessicaochen, line 34: (nit)
Should this also be like the other env vars GPU_CR......? It would be safer from conflict if it happens to be run in a shell with other processes that happen to also export files (with env var controls)

jessicaochen (dismissed):

pkg/snapshot-agent/backends/memory_regions.go
Edwinhr716: No, since this is a GPU-CR native environment variable

pkg/snapshot-agent/backends/memory_regions_test.go
Edwinhr716: Fixed

jessicaochen (approved):

PR #193, opened 2026-09-02
[Snapshot-Agent Memory-Regions 4/N]: Python Client & Configs

Edwinhr716:
## What does this PR do?

Adds `MemoryRegion` type to the Python Client, docs, and Configs to simplify usage (similar to #187 and #171)

## Why is this change needed?
Adds support for the MemoryRegions backend to the Python Client

## How was this tested?

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for selectively snapshotting and restoring explicit device-memory regions.
  * Added flexible region configuration using typed objects, tuples, or `pid:address:size` strings with decimal or hexadecimal values.
  * Added optional snapshot slot naming, defaulting to the job ID.
  * Exposed memory-region configuration tools through the Python client.
* **Bug Fixes**
  * Added validation for process IDs, addresses, sizes, snapshot names, and malformed region specifications.
* **Documentation**
  * Added usage guidance and examples for configuring memory-region snapshots.

### DIFF
--- a/pkg/client/python/tests/test_configs.py
+++ b/pkg/client/python/tests/test_configs.py
@@ -3,10 +3,12 @@
 import unittest
 
 from timeslice.snapshot_agent import (
+    MemoryRegion,
     app_channel_config,
     app_endpoint_config,
     cuda_config,
     direct_memory_config,
+    memory_regions_config,
     sglang_config,
     vllm_config,
 )
@@ -60,6 +62,19 @@ class TestBackendConfigBuilders(unittest.TestCase):
                 cuda_config(pids)
             with self.assertRaises((ValueError, TypeError)):
                 direct_memory_config(pids)
+
+    def test_memory_regions_config_builds_regions(self):
+        """Regions and snapshot_name land in the memory_regions oneof."""
+        cfg = memory_regions_config(
+            [MemoryRegion(pid=123, address=0x7F00, size_bytes=1024)],
+            snapshot_name="slot-a",
+        )
+        self.assertEqual(cfg.WhichOneof("backend"), "memory_regions")
+        self.assertEqual(cfg.memory_regions.snapshot_name, "slot-a")
+        region = cfg.memory_regions.regions[0]
+        self.assertEqual(region.pid, 123)
+        self.assertEqual(region.address, 0x7F00)
+        self.assertEqual(region.size_bytes, 1024)
 
     def test_app_endpoint_config_builds_full_config(self):
         cfg = app_endpoint_config(
--- a/pkg/client/python/tests/test_memory_regions.py
+++ b/pkg/client/python/tests/test_memory_regions.py
@@ -1,29 +1,40 @@
-import pytest
+"""Tests for the MemoryRegion type and the memory-regions config builder."""
+
+import unittest
 
 from timeslice.snapshot_agent import MemoryRegion, memory_regions_config
 from timeslice.snapshot_agent import snapshot_agent_pb2
 
 
-class TestMemoryRegionFromSpec:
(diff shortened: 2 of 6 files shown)


pkg/client/python/tests/test_configs.py
```diff
@@ -61,6 +63,19 @@ def test_rejects_non_integer_pids(self):
             with self.assertRaises((ValueError, TypeError)):
                 direct_memory_config(pids)
 
+    def test_memory_regions_config_builds_regions(self):
+        """Regions and snapshot_name land in the memory_regions oneof."""
+        cfg = memory_regions_config(
+            [MemoryRegion(pid=123, address=0x7F00, size_bytes=1024)],
+            snapshot_name="slot-a",
```
jessicaochen, line 70: snapshot_name is planned to be used as a directory name. Can we test somewhere that things that would be malformed or poor directory names are rejected?

(eg. backslash, forward slash, colon, whitespace).

Restricting it to letters,numbers,"-" & "_" is probably a good call.

pkg/client/python/timeslice/snapshot_agent/configs.py
```diff
@@ -164,3 +167,54 @@ def app_channel_config(
...
+                )
+            r = MemoryRegion(pid=region[0], address=region[1], size_bytes=region[2])
+        else:
+            raise TypeError(
+                "memory region must be a MemoryRegion, (pid, address, size_bytes) "
+                f"tuple, or 'pid:0xADDR:size' string, got {type(region).__name__}"
+            )
+        pb_regions.append(
+            snapshot_agent_pb2.MemoryRegion(
+                pid=r.pid, address=r.address, size_bytes=r.size_bytes
```
jessicaochen, line 229: A bit of a design question: how do we expect the flow to look like on k8s? IIRC, the agent has typically discovered the pid from the labels and then provides it to the underlying commands. What is the plan for memory regions?

I recall you had a different PR where you needed to make pids not required for several backends.

Edwinhr716: Addressed in the design. Agent won't discover pids, so we need to make pids required for this backend

pkg/client/python/tests/test_configs.py
Edwinhr716: Added check with regex for those, and tests to validate that invalid dir names are rejected.

pkg/client/python/tests/test_memory_regions.py
```diff
@@ -142,6 +142,35 @@ def test_snapshot_name_defaults_to_empty(self):
...
+    def test_accepts_safe_snapshot_names(self):
+        """Letters, digits, '-' and '_' are the allowed slot-name charset."""
+        for name in ("slot-a", "Slot_B", "0", "a-b_c-42"):
+            cfg = memory_regions_config([(1, 0x10, 16)], snapshot_name=name)
+            self.assertEqual(cfg.memory_regions.snapshot_name, name)
+
+    def test_rejects_unsafe_snapshot_names(self):
+        """snapshot_name becomes a directory on the agent: path separators,
+        traversal, and shell-hostile characters are rejected client-side."""
+        for name in (
```
jessicaochen, line 154: Since the documenation calls it out, add trailng, leading and all whitespace rejections

jessicaochen (dismissed):

pkg/client/python/tests/test_memory_regions.py
Edwinhr716: added

jessicaochen (approved):

Issue #155, opened 2026-08-18
[Feature]: Support multi host concurrent restore


lynnl0927:
### Problem Statement

Currently the controller issues per-node restore serially and blocks per operation, which breaks multi-node workloads. Any backend whose multi-node (single slice across multiple nodes) restore requires all nodes' operations in flight simultaneously (rendezvous semantics).

### Proposed Solution

Collect all SAVED nodes, issue every async Restore, then await all operations concurrently.

### Willingness to Contribute

Yes, I can submit a PR

PR #195, opened 2026-09-08
Support concurrent TPU checkpoint

lynnl0927:
## What does this PR do?

Support concurrent TPU checkpoint

## Why is this change needed?

Concurrent checkpoint across nodes is required for multi-node TPU training.

## How was this tested?

- [x] Integration/e2e tests added/updated

## Checklist

- [x] Tests pass locally (`make test`)
- [x] Linters pass (`make lint`)

Integration test results
  === RUN   TestOrchestrator
  === RUN   TestOrchestrator/SingleRLJob
      scenarios.go:41: Starting Single RL Job Scenario
      scenarios.go:100: Single RL Job Scenario completed successfully
  === RUN   TestOrchestrator/QueuedRLJobs (already failed test)
      fake_rl_job.go:108: [Job job-a] WARNING: Acquire for group trainers returned transient error
          (rpc error: code = Unavailable desc = group trainers is faulted), retrying in 500ms (attempt 20/20)...
      scenarios.go:185: [Test] Job A exited with error: loop failed: failed to acquire lock for trainers after retries
      orchestrator_test.go:86: QueuedRLJobs scenario failed: job A failed
  === RUN   TestOrchestrator/MultiNodeGroups
      scenarios.go:283: Starting Multi-Node Groups Scenario
  panic: Fail in goroutine after TestOrchestrator/QueuedRLJobs has completed
  FAIL    github.com/llm-d-incubation/llm-d-rl-time-slicing/tests/integration/orchestrator  436.338s

## Related Issues

https://github.com/llm-d-incubation/llm-d-rl-time-slicing/issues/155

## Summary by CodeRabbit

- **New Features**
  - Multi-node checkpoint snapshots and active-job restores now run concurrently across all nodes, supporting coordinated accelerator workloads.
  - Added configurable settling time before queued jobs are promoted.
  - Integration environments can optionally configure a second sampler node for multi-node scenarios.

- **Bug Fixes**
  - Restore failures now propagate correctly while ensuring all node operations are initiated before retry.
  - Missing active jobs are handled explicitly during restore operations.

- **Tests**
  - Added coverage for multi-node switching, rendezvous behavior, queued jobs, failure handling, and cleanup.

### DIFF
--- a/pkg/timeslice-orchestrator/controller/controller.go
+++ b/pkg/timeslice-orchestrator/controller/controller.go
@@ -260,11 +260,31 @@ func (c *Controller) reconcileGroup(ctx context.Context, groupID string) error {
 	activeJob := group.Spec().ActiveJob()
 
 	// 3. Act
-	// TODO: add optional fan out parallelism for node reconciliation
+	// 3a. Per-node early exits (active job already running / faulted / transitioning).
 	for _, node := range group.Status().Nodes() {
 		if err := c.reconcileNode(ctx, group.ID(), node, activeJob); err != nil {
 			return fmt.Errorf("failed to reconcile node %s: %w", node, err)
 		}
+	}
+
+	// 3b. Park (snapshot) any OTHER job that still has context loaded, fanning
+	// the snapshots out CONCURRENTLY across all nodes. A healthy host's snapshot
+	// is per-host independent, but issuing serially and blocking per node lets a
+	// single host's checkpoint stall freeze the whole group pass, delay the peer
+	// hosts' snapshots, and cascade into the waiting job's restore faulting.
+	if err := c.reconcileOtherJobsSnapshot(ctx, group, activeJob); err != nil {
+		return fmt.Errorf("failed to snapshot non-active jobs: %w", err)
+	}
+
+	// 3c. Restore the active job onto every node whose context is SAVED, fanning
+	// the restores out CONCURRENTLY. On multi-host accelerator slices (e.g. TPU
+	// libtpu) restore is a slice-wide rendezvous: every host's restore must be
+	// in flight simultaneously or none complete. Issuing them one node at a time
+	// and blocking on each deadlocks a multi-host restore — the first host parks
+	// in the rendezvous waiting for peers that are never triggered, until it
+	// times out and the job FAULTS.
+	if err := c.reconcileActiveJobRestore(ctx, group, activeJob); err != nil {
+		return fmt.Errorf("failed to restore active job: %w", err)
 	}
 
 	// 4. Update Status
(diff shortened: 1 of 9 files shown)


pkg/timeslice-orchestrator/controller/controller.go
```diff
@@ -238,13 +239,34 @@ func (c *Controller) reconcileGroup(ctx context.Context, groupID string) error {
 	activeJob := group.Spec().ActiveJob()
 
 	// 3. Act
-	// TODO: add optional fan out parallelism for node reconciliation
+	// 3a. Per-node early exits (active job already running / faulted / transitioning).
 	for _, node := range group.Status().Nodes() {
 		if err := c.reconcileNode(ctx, group.ID(), node, activeJob); err != nil {
```
jessicaochen: High level: I intended for the reconcileNode to be called in parallel when parallelization gets added. After all, the snapshot then restore should be sequential on a particular node but should not need to be sequential across the entire group. Is there some context I am missing?

jessicaochen (dismissed):

pkg/timeslice-orchestrator/controller/controller.go
lynnl0927: Do you mean instead of group-wide per-node checks -> snapshot fanout -> restore fanout, we do per-node fanout (snapshot-then-restore per node)?

In TPU multi-host slice, each process in the slice would park in the libtpu rendezvous until every peer's c/r is also in flight. So the restore step would be blocked anyways until all nodes in the slice have successfully finished their snapshot. 

Technically both per-node and per-phase fanout works, it's just if one node finishes snapshotting earlier (or somehow has nothing to snapshot due to some corner cases), it'll have to wait for other nodes to finish snapshotting while it parks in restore rendezvous, which eats the restore timeout budget. 

I can change to per-node fanout if you'd prefer that? The code structure would look cleaner for per-node fanout.

jessicaochen: I indeed would prefer per node fanout. It would be easier to read and follow.

pkg/timeslice-orchestrator/controller/controller.go
```diff
@@ -259,12 +260,53 @@ func (c *Controller) reconcileGroup(ctx context.Context, groupID string) error {
...
+			defer wg.Done()
+			defer func() {
+				if r := recover(); r != nil {
+					slog.ErrorContext(ctx, "Observed a panic", "panic", r, "stack", string(debug.Stack()))
+					nodeErrs[idx] = fmt.Errorf("panic while reconciling node %s: %v", node, r)
+					cancelPeers()
+				}
+			}()
+			if err := c.reconcileNode(opCtx, group.ID(), node, activeJob, cancelPeers); err != nil {
+				nodeErrs[idx] = fmt.Errorf("failed to reconcile node %s: %w", node, err)
```
aishukamal, line 303: coderabbit has a valid point here. Is there a reason we choose to `cancelPeers()` only on panic and not other failure cases?

aishukamal (approved):

